PR #59, opened 2026-05-13
Add IO-level scheduler — squad-independent recurring tasks (#58)

### LINKED ISSUE #58: Add native IO task scheduling without requiring a squad
## Problem

The scheduler (squad_schedule_create) is squad-only. There is no way to schedule recurring tasks for IO itself without creating a squad as a workaround.

## Proposed Solution

Add a native scheduling system for IO that allows standalone recurring tasks independent of squads. This could follow a similar cron-based interface to squad schedules.

## Example Use Cases

- Daily reminders or digests
- Periodic health checks
- Recurring research or monitoring tasks
- Any automation that does not belong to a specific project squad

## Proposed API

- schedule_create: create a recurring IO task with a cron expression and prompt
- schedule_list: list all IO-level schedules
- schedule_pause / schedule_resume / schedule_delete: lifecycle management

## Notes

Current workaround is a system cron job calling the IO HTTP API, but this is brittle and not manageable through the IO interface.

michaeljolley:
Closes #58.

## Summary
Native scheduling for IO itself — no squad required. Mirrors the squad scheduler in shape but dispatches into the orchestrator via `sendToOrchestrator` with a `background` source, so a scheduled task is handled the same way as a TUI/Telegram message.

## New tools
- `schedule_create` — name + cron + prompt + optional notes
- `schedule_list`
- `schedule_pause` / `schedule_resume`
- `schedule_delete`
- `schedule_run_now` — manual fire for testing

## Storage
New `io_schedules` table: id, name, cron_expr, prompt, notes, enabled, last_run_at, next_run_at, created_at. Indexed on (enabled, next_run_at) to match the squad-schedules pattern.

## Scheduler
`src/copilot/io-scheduler.ts` — 30s tick, in-flight guard, advances `next_run_at` before dispatch (no double-fires), startup reconcile that skips missed runs (same semantics as squad scheduler). Wired into `src/daemon.ts` start/stop.

## System message
Adds an "IO-level Schedules (squad-independent)" section so the orchestrator reaches for `schedule_create` on prompts like "remind me every morning at 9AM to check the dashboard".

## Verification
Build clean. End-to-end CRUD + reconcile smoke test on a fresh DB:
- create → enabled, next_run_at set
- pause → enabled=0
- resume + reconcile → next_run_at advanced
- delete → removed

🦁

### DIFF
--- a/src/copilot/io-scheduler.ts
+++ b/src/copilot/io-scheduler.ts
@@ -8,6 +8,7 @@ import {
   listIoSchedules,
   listDueIoSchedules,
   recordIoScheduleRun,
+  setIoScheduleTimestamps,
   updateIoScheduleNextRun,
   type IoSchedule,
 } from "../store/io-schedules.js";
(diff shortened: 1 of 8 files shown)


michaeljolley:
Pushed fix for the run_now bug 🦁

The reviewer correctly flagged the misleading JSDoc on `runIoScheduleNow`. On investigation, **both** `runIoScheduleNow` and `runScheduleNow` (squad scheduler) had the same bug — the comment was the symptom, the behaviour was the actual problem. `fireSchedule` always advances `last_run_at` and `next_run_at` as a side effect, so a user testing a 05:00 schedule at 04:30 would silently skip today's 05:00 occurrence.

**Fix** (commit c472632):
- Snapshot `last_run_at` and `next_run_at` before `fireSchedule`, restore in `finally` so the schedule is left pristine even if the dispatch throws.
- New `setIoScheduleTimestamps` / `setScheduleTimestamps` store helpers to support the NULL `last_run_at` case (never-run schedules).
- Same fix applied to the squad scheduler for consistency.
- Tool descriptions updated to accurately describe the behaviour: _"last_run_at and next_run_at are preserved so the regular schedule is untouched."_

**Verification:** end-to-end smoke test confirms a manual fire on a freshly-created schedule leaves both timestamps unchanged (NULL → NULL, future → future). Build clean.

PR #91, opened 2026-05-13
Fix: activity panel fallback for tasks after daemon restart (#66)

### LINKED ISSUE #66: Activity panel shows "no activity" for tasks that have data
## Problem

When viewing a task's history in the web UI, the activity panel shows "no activity" even when the task has run. Toggling "Show Details" on/off has no effect — the panel remains empty.

## Steps to Reproduce

1. Navigate to the activity/history view for a task
2. Observe the activity panel shows "no activity"
3. Click "Show Details" toggle — no change

## Expected Behavior

The activity panel should display the task's activity log. The "Show Details" toggle should expand/collapse additional detail.

## Additional Context

Reproduced with task ID  (status: failed). The task did run and produce output (timed out after 600s), but the UI shows nothing.

michaeljolley:
Closes #66.

## Problem
Task events are stored in-memory only. After a daemon restart, the activity panel showed 'no activity' for tasks that had data in SQLite.

## Fix
When in-memory events are empty, the `GET /tasks/:taskId/activity` endpoint now falls back to a synthetic `ActivityEntry` built from the persisted `result` field in the `agent_tasks` table. The entry is clearly labelled as `(activity log unavailable after restart)` and uses `kind: 'outcome'` with the appropriate status icon.

## Scope
Single-file change: `src/api/server.ts`. No frontend changes needed — the UI already handles the activity array correctly.

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -244,7 +244,28 @@ export async function startApiServer(): Promise<void> {
     const taskId = Array.isArray(req.params.taskId) ? req.params.taskId[0] : req.params.taskId;
     try {
       const events = getTaskEvents(taskId);
-      const activity = summarize(events);
+      let activity = summarize(events);
+
+      // Fallback: when in-memory events are gone (e.g. daemon restart),
+      // build a minimal entry from the persisted task result so the UI
+      // doesn't show "no activity" for tasks that actually ran. (#66)
+      if (activity.length === 0) {
+        const task = getTask(taskId);
+        if (task?.result) {
+          activity = [{
+            ts: task.completed_at ? new Date(task.completed_at).getTime() : Date.now(),
+            kind: "outcome" as const,
+            icon: task.status === "failed" ? "\u274c" : task.status === "done" ? "\u2705" : "\ud83d\udccb",
+            summary: task.status === "failed"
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

Reviewed diff, `gh issue view 66`, ran `npx tsc --noEmit` (clean) and `npm test` (56/56 pass on branch).

**Correctness** — Fallback fires only when `activity.length === 0` (in-memory events cleared after restart). Existing tasks with live events are completely unaffected — the fallback block is never entered when `activity.length > 0`.

**Type safety** — Synthetic entry satisfies the full `ActivityEntry` interface: `ts`, `kind: "outcome"`, `icon`, `summary`, `rawType`, `raw` all present. Optional `detail` is populated. Optional `toolCallId`/`status` correctly omitted.

**Edge cases all handled:**
- `getTask(taskId)` returns `undefined` → `task?.result` is `undefined` → falsy → skipped → `activity: []` returned. Safe.
- `task.result === null` → falsy → skipped. Correct — a task with no stored result gets no synthetic entry.
- `task.completed_at === null` → `Date.now()` fallback. Safe for orphaned/running tasks.

**Approve** — this is a minimal, correct bugfix with no regressions.

PR #92, opened 2026-05-13
Agent preview defaults to human-readable summary view (#67)

### LINKED ISSUE #67: Agent preview in Squads section should default to human-readable view
## Feature Request

When clicking the "preview" button for a running agent in the Squads section of the web UI, the view should default to a human-readable summary — consistent with the tiered logging experience described in #42.

## Expected Behavior

- **Default view**: Human-readable summary (assistant messages, tools used, reasoning, outcomes)
- **Detail view**: Full raw output accessible via a toggle or drill-down

## Current Behavior

The preview shows raw output with no human-readable default.

## Notes

This should be consistent with how activity history is displayed. The detail toggle behavior should mirror whatever is implemented for the history view in #42.

michaeljolley:
Closes #67.

The live agent preview in the Squads section now defaults to a summary mode showing human-readable activity entries (icon + summary + timestamp) instead of raw SSE JSON. A Summary toggle lets users switch to the raw view.

Reuses the server-side ActivityEntry payload already attached to SSE events — zero client-side duplication.

Single file: web/src/views/SquadsView.vue.

### DIFF
--- a/web/src/views/SquadsView.vue
+++ b/web/src/views/SquadsView.vue
@@ -187,12 +187,18 @@
               {{ previewStatusLabel }}
             </p>
           </div>
-          <button
-            type="button"
-            @click="closePreview"
-            class="text-gray-400 hover:text-gray-100 text-2xl leading-none focus:outline-none"
-            aria-label="Close"
-          >×</button>
+          <div class="flex items-center gap-3">
+            <label class="flex items-center gap-1.5 text-xs text-gray-400 cursor-pointer select-none">
+              <input type="checkbox" v-model="previewSummaryMode" class="rounded bg-gray-700 border-gray-600" />
+              Summary
+            </label>
+            <button
+              type="button"
+              @click="closePreview"
+              class="text-gray-400 hover:text-gray-100 text-2xl leading-none focus:outline-none"
+              aria-label="Close"
+            >×</button>
+          </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

Verified `npx tsc --noEmit` clean and `npm test` 56/56 pass on branch. Reviewed the diff against issue #67.

**Correctness** ✓
- `previewSummaryMode` initialises to `true` and resets to `true` on every `openPreview()` call — defaults to summary as required.
- The SSE endpoint (`/api/tasks/:taskId/events`) already wraps every event with `{ ...ev, summary: summarizeEvent(ev) }` server-side, so `PreviewEvent.summary` is always populated (both replayed and live events). The `!ev.summary` guard is defensive but safe.
- Raw mode is the original block, byte-for-byte unchanged.

**Vue practices** ✓ — scalar `ref(true)`, derived `computed()`, `v-model` checkbox, `v-else-if` / `v-else` template blocks. No anti-patterns.

**Edge cases** ✓
- Summary produces zero visible events (e.g. only system/delta events so far) → "Processing…" shown, not a blank.
- "Waiting for activity…" outer guard fires before either content block when `previewEvents` is empty — order is correct.
- `assistant.message_delta` filtered out (avoids flooding the summary with partial tokens). 

**Minor UX note** (not a blocker): the checkbox + "Summary" label is easy to miss — a toggle switch or small segmented control would be more discoverable. Worth a follow-up polish issue.

**Approve** — implementation is correct, matches issue requirements, no regressions.

PR #93, opened 2026-05-13
Multi-line chat input with auto-grow textarea (#81)

### LINKED ISSUE #81: Web UI chat input should support multi-line text
## Feature Request

The chat input box in the web UI is currently a single-line input field. It should support multi-line input so users can compose longer, structured messages (e.g., pasting code, writing multi-paragraph prompts, or formatting task instructions).

## Expected Behavior

- The input expands vertically as the user types (auto-grow textarea)
-  inserts a newline;  alone submits
- Minimum height stays compact for short messages; max height caps with scrolling for very long input

michaeljolley:
Closes #81.

Replaces the single-line input with an auto-growing textarea. Enter submits, Shift+Enter inserts a newline. Height auto-grows up to ~8 lines and resets after sending.

Single file: web/src/views/ChatView.vue.

### DIFF
--- a/web/src/views/ChatView.vue
+++ b/web/src/views/ChatView.vue
@@ -37,13 +37,16 @@
 
     <!-- Input bar -->
     <div class="border-t border-gray-800 p-4">
-      <form @submit.prevent="sendMessage" class="flex gap-3">
-        <input
+      <form @submit.prevent="sendMessage" class="flex gap-3 items-end">
+        <textarea
+          ref="inputEl"
           v-model="input"
-          type="text"
           placeholder="Message IO..."
           :disabled="store.isLoading"
-          class="flex-1 bg-gray-800 border border-gray-700 rounded-lg px-4 py-2 text-sm text-gray-100 placeholder-gray-500 focus:outline-none focus:border-blue-500 disabled:opacity-50"
+          rows="1"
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Auto-grow** ✓ — `height = 'auto'` → `Math.min(scrollHeight, 200)px` is the correct reset-then-measure pattern. `resize-none overflow-hidden` prevents the native handle from conflicting.

**Keyboard handling** ✓ — `Enter` alone calls `sendMessage()` + `preventDefault()`; `Shift+Enter` falls through to browser default newline. Matches issue spec exactly.

**Reset on send** ✓ — `input.value = ''` clears the model; `nextTick(() => el.style.height = 'auto')` fires after Vue re-renders so the textarea collapses correctly.

**No regressions** ✓ — `sendMessage()` logic untouched; `v-model`, `:disabled`, form submit path all preserved. `nextTick` was already imported.

**Minor UX note** (not a blocker): `overflow-hidden` at the 200px cap clips content rather than scrolling. Switching to `overflow-y: auto` when at max height would let users scroll/read their full paste. Worth a follow-up polish issue but doesn't block merge.

**Approve** — implementation is correct and complete per issue #81.

PR #94, opened 2026-05-13
Scroll-to-bottom button in chat when scrolled up (#82)

### LINKED ISSUE #82: Web UI chat: show "scroll to bottom" button when scrolled up
## Feature Request

When a user is scrolled up in the chat window, a floating "scroll to bottom" button should appear in the bottom-right corner of the chat area. Clicking it instantly scrolls to the latest message.

## Expected Behavior

- Button is hidden when already at (or near) the bottom
- Button appears when the user scrolls up past a threshold
- Clicking scrolls smoothly (or instantly) to the most recent message
- A downward arrow icon (e.g.,  or a chevron) is sufficient — no label needed

michaeljolley:
Closes #82.

Floating chevron-down button appears when the user scrolls up in chat. Auto-scroll on new messages is suppressed while scrolled up. Sending a message always scrolls to bottom.

Single file: web/src/views/ChatView.vue.

### DIFF
--- a/web/src/views/ChatView.vue
+++ b/web/src/views/ChatView.vue
@@ -1,7 +1,7 @@
 <template>
-  <div class="flex flex-col h-full">
+  <div class="flex flex-col h-full relative">
     <!-- Message history -->
-    <div ref="messagesEl" class="flex-1 overflow-y-auto p-6 space-y-4">
+    <div ref="messagesEl" class="flex-1 overflow-y-auto p-6 space-y-4" @scroll="handleScroll">
       <div v-if="store.messages.length === 0" class="text-gray-600 text-sm text-center mt-12">
         Send a message to start chatting with IO
       </div>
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Correctness** ✓ — `isScrolledUp` computed from `scrollTop + clientHeight < scrollHeight - 100`. Button shown with `v-if`. Click calls the existing `scrollToBottom()`. Correct.

**Auto-scroll behavior** ✓ — `watch(messages.length)` now guards with `if (!isScrolledUp.value)`, so arriving messages don't force-scroll a reading user. User-initiated sends always scroll (explicit call after send) — correct, they want to see the response.

**Threshold** ✓ — 100px is standard and handles subpixel rounding without being too eager.

**Positioning** ✓ — `relative` added to outer container; button is `absolute bottom-20 right-8` so it floats within the chat area, not the viewport.

**Accessibility** ✓ — `title="Scroll to bottom"` tooltip, natively focusable button, keyboard-activatable.

**No regressions** ✓ — scroll listener is purely additive; `scrollToBottom`, `sendMessage`, and `isStreaming` logic unchanged.

**Minor note (not a blocker):** `handleScroll` fires on every raw scroll event with no debounce/throttle. For typical chat lengths this is fine, but wrapping in `requestAnimationFrame` would be a cheap guard against jank on large message lists. Worth a follow-up polish if perf ever shows up.

**Approve** — correct and complete per issue #82.

PR #95, opened 2026-05-13
Install skills from direct SKILL.md URLs (#84)

### LINKED ISSUE #84: Allow installing a skill from a specific SKILL.md path within a repo
## Problem

Currently, installing a skill requires providing the full repo URL — IO clones the entire repository and expects a  at the root. This is problematic when:

- A repo contains multiple skills (e.g. a skill library), but you only want one specific skill.
- You don't want the overhead of cloning a full repo just to grab a single skill definition.

## Proposed Solution

Allow  (and the underlying skill install flow) to accept a **path to a specific  file** within a repo, in addition to the existing full-repo URL.

Possible input formats to support:

- A GitHub file URL: 
- A raw GitHub URL: 
- A  shorthand

IO should fetch only that file (no clone required) and install the skill from it directly.

## Acceptance Criteria

- User can provide a path/URL pointing to a specific  file in a repo
- IO fetches only that file — no full repo clone needed
- Skill is installed and behaves identically to one installed via full repo URL
- Existing full-repo-URL install flow is unchanged
- Error message is clear if the path doesn't resolve to a valid 

## Notes

- This is especially useful for skill library repos where multiple skills live under subdirectories
- The  tool description and any relevant docs should be updated to reflect the new input format(s)

michaeljolley:
Closes #84.

`installSkill()` now accepts GitHub blob URLs, raw.githubusercontent URLs, and any HTTPS URL ending in SKILL.md. File-based installs fetch just the SKILL.md (no clone), validate it has a markdown heading, and derive a slug from the repo/path. Existing repo-URL flow is unchanged.

Files: `src/copilot/skills.ts`, `src/copilot/tools.ts`.

### DIFF
--- a/src/copilot/skills.ts
+++ b/src/copilot/skills.ts
@@ -1,4 +1,4 @@
-import { existsSync, readdirSync, readFileSync, rmSync, statSync } from "fs";
+import { existsSync, mkdirSync, readdirSync, readFileSync, rmSync, statSync, writeFileSync } from "fs";
 import { join, basename } from "path";
 import { execSync } from "child_process";
 import { SKILLS_DIR } from "../paths.js";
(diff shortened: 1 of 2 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Security** ✓
- GitHub blob/raw regexes are hardcoded `https://`; GENERIC path has an explicit `startsWith("https://")` guard that throws. `installSkillFromFile` has a second HTTPS check as defense in depth.
- No path traversal: slugs are composed of `[^\/]+` repo-name segments + a single `filter(Boolean)` path segment. `repo-..` is a safe literal filename, not interpreted as a path component by `join()`.

**Correctness** ✓
- Root-level SKILL.md (`pathPrefix = undefined`) → slug = `repo` ✓
- Nested path → last segment appended: `repo-myskill` ✓
- Blob→raw URL derivation correct for both named and root paths ✓
- `installSkill()` signature is unchanged; all callers work without modification ✓

**Error handling** ✓
- Non-200 response → throws with HTTP status
- 200 HTML error page → heading heuristic (`/^#\s+/`) catches it (HTML won't have an ATX heading in first 10 lines)
- Already-installed slug → clear error message

**Heading heuristic** ✓ — Requires `# heading` (space after `#`), which is the correct ATX syntax. Intentionally strict; fine.

**Integration** ✓ — API endpoint and tools handler both call `installSkill(str)`; dispatch is internal and transparent.

**Non-blocking notes:**
1. **Slug collision across orgs** — `user1/skills/.../mymod/SKILL.md` and `user2/skills/.../mymod/SKILL.md` both produce `skills-mymod`. The `existsSync` guard throws "already installed" rather than explaining the collision. Acknowledged as a known trade-off per squad decision (2026-05-13 20:55:14). Could be a follow-up to add owner to the slug for public library URLs.
2. **No unit tests for `parseSkillUrl`** — it's a pure function with 4 distinct branches (blob, raw, generic, repo fallback) and would be a quick high-value test addition.

**Approve** — implementation is correct, secure, and meets all acceptance criteria from issue #84.

PR #97, opened 2026-05-13
Schedule execution history table + daily pruning (#86)

### LINKED ISSUE #86: Schedule execution history table + periodic notification pruning
## Feature Request

Both `io-scheduler.ts` and `scheduler.ts` log fire events to stdout but there's no queryable execution history. If a schedule silently fails at 3 AM, the only evidence is buried in daemon logs.

### 1. `schedule_runs` table
Record each fire: `schedule_id, started_at, completed_at, status (success/error/timeout), error_text, notification_id (FK to background_notifications)`. This lets `squad_status` and the web UI show "last 5 runs" per schedule with pass/fail, and links a notification back to the run that produced it.

### 2. Periodic pruning sweep
`pruneOldNotifications(olderThanDays)` is implemented (PR #83) but never called. Add a daily timer in `daemon.ts` that prunes both `background_notifications` and `schedule_runs` older than a configurable retention period (default 30 days).

## Acceptance Criteria
- [ ] `schedule_runs` table with migration
- [ ] Both schedulers record a row per fire
- [ ] `squad_status` or schedule-list output shows last N runs per schedule
- [ ] Daily pruning sweep for old notifications and schedule runs

*Filed during daily stand-up ideation. Lion-O + WilyKit domains.*

michaeljolley:
Closes #86.

New `schedule_runs` table records every schedule fire with status, timing, and a link to the `background_notifications` row. Both schedulers (IO + squad) create a run record before dispatch and mark it done/failed on completion. Daily timer prunes runs and notifications older than 30 days.

Fan-out: WilyKit (store + migration) → Lion-O (scheduler wiring + daemon pruning timer).

### DIFF
--- a/src/copilot/io-scheduler.ts
+++ b/src/copilot/io-scheduler.ts
@@ -15,6 +15,7 @@ import {
 import { sendToOrchestrator } from "./orchestrator.js";
 import { nextRun } from "./cron.js";
 import { notifyBackground } from "../notify.js";
+import { startScheduleRun, completeScheduleRun, failScheduleRun } from "../store/schedule-runs.js";
 
 const TICK_MS = 30_000;
 
@@ -47,6 +48,11 @@ async function fireSchedule(schedule: IoSchedule): Promise<void> {
   console.log(
     `[io] io-scheduler: firing schedule "${schedule.name}" (next run: ${nextIso ?? "never"})`,
   );
+  const run = startScheduleRun({
+    schedule_type: "io",
+    schedule_id: schedule.id,
+    schedule_name: schedule.name,
+  });
   try {
     let buffer = "";
     await sendToOrchestrator(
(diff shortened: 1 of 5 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Schema** ✓ — `schedule_runs` table well-structured. Index on `(schedule_type, schedule_id, started_at)` matches the `getScheduleRuns` query exactly. `notification_id` is a soft FK (no REFERENCES constraint) — correct, prevents cascades when notifications are pruned.

**SQL** ✓ — `pruneOldScheduleRuns` uses `datetime('now', '-' || ? || ' days')`; when `?=30` this produces `'-30 days'`, a valid SQLite modifier. All queries parameterised. `startScheduleRun` uses `lastInsertRowid` for immediate read-back.

**Pruning** ✓ — Daily `setInterval(24h).unref()` won't block process exit. Startup `setTimeout(5s)` fires once. The cast on `pruneStartup.unref()` is overly defensive (TypeScript knows the return type has `.unref()`) but harmless.

**No regressions** ✓ — Both schedulers chain `.then/.catch` onto the existing `void notifyBackground()` call, preserving fire-and-forget semantics.

**`notifyResult.id` when skipped** ✓ — When output is skipped ("not-meaningful", "empty"), `notifyResult.id` is `undefined`; `completeScheduleRun(id, undefined)` resolves to `NULL` via `notificationId ?? null`. No crash; run still recorded as `done`.

**Two non-blocking notes:**

1. **Run stays `running` until notification resolves** — `completeScheduleRun` is called inside `notifyBackground().then()`. If Telegram send is slow, the run row stays `running` after the actual schedule work is done. Completing the run immediately after `delegateToAgent` returns and optionally linking the notification ID later would be more accurate. Cosmetic issue only.

2. **No unit tests for `schedule-runs.ts`** — All 5 helpers are DB-only with no external deps and would be a quick addition using the established `setDbPathForTests` pattern. Third time I've flagged this module; should be added before the feature gets built on further.

**Approve** — correct, schema is solid, pruning is properly wired, no regressions.

PR #100, opened 2026-05-13
Display version number in nav footer (#90)

### LINKED ISSUE #90: Display current version number near logged-in user in the UI
## Problem

There is currently no way to tell which version of IO is running from within the UI itself. Users have to check externally (e.g. package.json or GitHub releases) to know what version they're on.

## Proposed Solution

Display the current release version number in the bottom-left area of the UI, near where the logged-in user's name is shown.

**Example:**
```
Michael Jolley  v0.6.0
```

The version should be read from the running application (e.g. from `package.json` version field or a build-time constant) so it always reflects the actual deployed version.

## Acceptance Criteria

- Version number is displayed in the bottom-left UI area alongside the logged-in user
- Version reflects the current release (not hardcoded)
- Styling is subtle / secondary so it doesn't compete with the user's name

michaeljolley:
Closes #90.

Shows the running IO version (fetched from GET /api/status) in the nav sidebar footer, next to the user email. Subtle text-[10px] styling. Shown regardless of auth state.

Single file: web/src/components/AppNav.vue.

### DIFF
--- a/web/src/components/AppNav.vue
+++ b/web/src/components/AppNav.vue
@@ -55,28 +55,48 @@
       </ul>
     </div>
 
-    <div v-if="auth.authEnabled && auth.user" class="p-4 border-t border-gray-800">
-      <p class="text-xs text-gray-500 truncate mb-2">{{ auth.user.email }}</p>
-      <button
-        @click="handleSignOut"
-        class="w-full px-3 py-1.5 text-sm text-gray-400 hover:text-white hover:bg-gray-800 rounded transition-colors"
-      >
-        Sign out
-      </button>
+    <div class="p-4 border-t border-gray-800">
+      <template v-if="auth.authEnabled && auth.user">
+        <div class="flex items-center justify-between mb-2">
+          <p class="text-xs text-gray-500 truncate">{{ auth.user.email }}</p>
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Data source** ✓ — `/api/status` returns `{ version: IO_VERSION }` where `IO_VERSION` comes from `package.json` at startup — not hardcoded.

**Error handling** ✓ — `try/catch` with silent best-effort. On failure `version.value` stays `''` and `v-if="version"` hides the span. No broken UI on daemon restart or auth failure.

**Correctness** ✓ — Two render paths covered: auth-enabled-with-user (flex row: email + version on one line) and auth-disabled/unauthenticated (`v-else-if` shows version alone). Both correct.

**Styling** ✓ — `text-[10px] text-gray-600 shrink-0` is appropriately subtle. `shrink-0` prevents compression when email addresses are long.

**No regressions** ✓ — Outer footer `div` is now always rendered (holds version even when logged out). Auth-gated content moved to inner `<template v-if>`. Correct restructure.

**Bonus:** `/api/status` sits above the `requireAuth` middleware, so the version fetch succeeds even on the login page — version shows consistently regardless of auth state. ✓

**Approve** — correct, handles all states, appropriately subtle styling.

PR #104, opened 2026-05-13
feat: filter schedule history to show only relevant entries (#65)

### LINKED ISSUE #65: Filter schedule history to show only relevant entries
## Problem

When clicking on the history of a schedule, it currently shows all history entries. This makes it difficult to find entries specific to a particular schedule (e.g., a daily standup).

## Proposed Solution

The history view should be filtered to only show entries that belong to the selected schedule. For example, clicking the history of a daily standup schedule should show only that schedule's run history, not all history across all schedules.

## Acceptance Criteria

- [ ] Clicking history on a schedule navigates to a filtered history list
- [ ] The filtered list only shows history entries for the selected schedule
- [ ] The filtered view clearly indicates which schedule's history is being shown

michaeljolley:
Closes #65

## Changes
- **API**: Added `GET /schedules/:type/:id/runs?limit=N` endpoint returning filtered schedule run history from the `schedule_runs` table (added in #97)
- **Frontend**: Replaced RouterLink history buttons with inline expandable history panels in SchedulesView.vue. Each schedule now shows its own run history with status icons, timestamps, and error details.

### How it works
- Clicking 📋 History toggles an inline panel below the schedule card
- First open fetches from the API; subsequent opens use cached data
- Each run shows ✅ done / ❌ failed / ⏳ running status, start/completion times, and error text
- Panel header shows which schedule's history is displayed

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -13,6 +13,7 @@ import { IO_VERSION } from "../paths.js";
 import { requireAuth } from "./auth.js";
 import { listSchedules, getSchedule, deleteSchedule, setScheduleEnabled } from "../store/schedules.js";
 import { listIoSchedules, getIoSchedule, deleteIoSchedule, setIoScheduleEnabled } from "../store/io-schedules.js";
+import { getScheduleRuns } from "../store/schedule-runs.js";
 import { runScheduleNow } from "../copilot/scheduler.js";
 import { runIoScheduleNow } from "../copilot/io-scheduler.js";
 import {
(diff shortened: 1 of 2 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**API** ✓
- URL type `squads` → DB type `squad` via `scheduleTypeMap`. Unknown type → 400 with clear message. Non-numeric ID → 400. Limit defaulted to 25, capped at 100. Error handling with `try/catch` and 500 fallback. All correct.

**Frontend** ✓
- `toggleHistory` creates new `Map`/`Set` instances on every mutation — correct Vue reactivity pattern (in-place mutation won't trigger updates).
- Fetch-on-first-open: `historyData.has(key)` guard prevents redundant API calls on re-open after successful load.
- `finally` block always clears the loading state — no stuck spinner on error.
- `fetchSchedules()` clears `historyData` cache on refresh — new runs picked up correctly.

**Type safety** ✓ — `ScheduleRun` interface exactly matches the API response shape. `ns: 'io' | 'squads'` typed correctly. No `as any`.

**Edge cases** ✓
- Empty history → "No history yet" message ✓
- `error_text` → red text displayed ✓
- In-progress (`status='running'`) → ⏳ + gray badge ✓  
- `completed_at === null` → completion line hidden via `v-if` ✓

**No regressions** ✓ — Existing pause/resume/delete/run-now unchanged. Old `RouterLink` to `/activity?schedule=...` (the broken filter) correctly replaced with inline toggle.

**Minor note (non-blocking):** On API failure, `historyData` is never set for that key, so re-opening will retry the fetch (shows spinner again). This is acceptable behaviour — just means a persistent backend failure gives a "retry on every toggle" UX rather than a permanent error state.

**Approve** — correct implementation of issue #65, all edge cases handled, no regressions.

PR #107, opened 2026-05-14
fix: schedule card overflow with expand/collapse for long content (#101)

### LINKED ISSUE #101: Squad schedule cards overflow causing horizontal scrollbar
## Problem

Squad schedule cards don't wrap their content, causing long text to overflow horizontally and produce an unwanted scrollbar on the page.

## Proposed Solution

Redesign the schedule cards to handle overflow gracefully. Options to consider:

- **Expandable/collapsible cards** — show a truncated summary by default with a "expand" toggle to reveal full details
- **Text wrapping** — ensure all card content wraps within the card boundaries
- **Responsive layout** — cards reflow/resize based on available width

An expand/collapse pattern would also improve the overall UX by keeping the schedule list scannable while still making full details accessible.

## Acceptance Criteria

- [ ] No horizontal scrollbar appears when schedule cards contain long content
- [ ] All schedule card content is accessible (not clipped or hidden permanently)
- [ ] Cards remain visually consistent with the rest of the UI

michaeljolley:
Closes #101

## Changes
- Added overflow constraints on schedule card containers
- Action buttons wrap instead of forcing horizontal scroll
- Long schedule names and cron expressions break properly
- Prompt/agenda text truncated to 2 lines with Show more/less toggle for long content
- History run rows wrap on narrow viewports

### DIFF
--- a/web/src/views/SchedulesView.vue
+++ b/web/src/views/SchedulesView.vue
@@ -37,30 +37,31 @@
             <div
               v-for="s in ioSchedules"
               :key="s.id"
-              class="bg-gray-800 border border-gray-700 rounded-lg p-4 hover:border-gray-600 transition-colors"
+              class="bg-gray-800 border border-gray-700 rounded-lg p-4 hover:border-gray-600 transition-colors overflow-hidden"
             >
-              <div class="flex flex-wrap items-start justify-between gap-3">
+              <!-- Top row: info + actions -->
+              <div class="flex flex-wrap items-start gap-3">
                 <!-- Info -->
                 <div class="flex-1 min-w-0">
-                  <div class="flex items-center gap-2 mb-1">
-                    <span class="font-bold text-gray-100">{{ s.name }}</span>
+                  <div class="flex flex-wrap items-center gap-2 mb-1">
+                    <span class="font-bold text-gray-100 break-words min-w-0">{{ s.name }}</span>
                     <span :class="[
-                      'px-2 py-0.5 rounded text-xs font-medium',
+                      'px-2 py-0.5 rounded text-xs font-medium shrink-0',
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**Overflow fix** ✓ — Three correct techniques in combination: `overflow-hidden` on card root, `break-words` on name/notes, `break-all` on cron expression (no natural break points), `min-w-0` on flex info child (prevents flex blowout). Covers all the ways long content can cause horizontal scroll.

**Expand/collapse** ✓ — `togglePrompt` creates a new `Set` on each mutation — correct Vue reactivity. Toggle button gated on `s.prompt.length > 120`. `line-clamp-2` collapses; removing the class expands. Clean.

**Short content / no prompt** ✓ — `v-if="s.prompt"` gates the entire block; `v-if="s.prompt.length > 120"` gates the button. Cards with short or no prompts are completely unaffected.

**Action buttons** ✓ — Old `justify-between` + `shrink-0` on actions would push buttons off-screen on narrow cards. Replaced with `flex flex-wrap` so buttons wrap instead of overflowing. Correct.

**History run rows** ✓ — Same `flex-wrap` + `shrink-0` on status badge applied consistently to the history panel too.

**Minor note (non-blocking):** The 120-char toggle threshold is a character count, but `line-clamp-2` is CSS-based (depends on font/container width). A <120-char prompt could still overflow 2 lines at narrow widths and show no toggle. In that case `break-words` still prevents horizontal overflow — it just means the text is clipped at 2 lines with no way to expand. Acceptable edge case; would need a ResizeObserver-based approach to fix perfectly, which would be over-engineering here.

**Approve** — correct and complete fix for issue #101. All overflow vectors addressed.

PR #108, opened 2026-05-14
fix: persist auth session across page refresh (#102)

### LINKED ISSUE #102: Refreshing the web UI logs the user out and redirects to login screen
## Problem

Refreshing the web UI (F5 or hard refresh) redirects to the login screen instead of keeping the user logged in. The authenticated session is not being persisted across page loads.

## Expected Behavior

User should remain logged in after a browser refresh.

## Acceptance Criteria

- [ ] Refreshing the page does not redirect to the login screen
- [ ] Authenticated session persists across page reloads

michaeljolley:
Closes #102

## Changes
- **Dedup concurrent init()**: Added promise-based guard so multiple router.beforeEach calls during refresh share one init
- **Wait for INITIAL_SESSION**: Register onAuthStateChange before getSession() and resolve on Supabase v2's INITIAL_SESSION event, which fires exactly once with the fully-settled auth state from localStorage

## Root Cause
On page refresh, the router guard called auth.init() which called getSession() before Supabase finished restoring the session from localStorage. The result was null, triggering a redirect to login.

### DIFF
--- a/web/src/stores/auth.ts
+++ b/web/src/stores/auth.ts
@@ -10,7 +10,10 @@ export const useAuthStore = defineStore('auth', () => {
   const initialized = ref(false)
   const authEnabled = ref(false)
 
-  async function init() {
+  // Deduplicates concurrent init() calls during page load / multi-navigation on refresh
+  let initPromise: Promise<void> | null = null
+
+  async function doInit() {
     const config = await getAuthConfig()
     authEnabled.value = config.authEnabled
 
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** ✅

**`initPromise` dedup** ✓ — Three-layer guard: (1) `initialized.value` fast-path for already-done calls, (2) `initPromise` singleton ensures only one `doInit()` ever runs, (3) `initialized.value = true` at end of `doInit` so all future calls short-circuit immediately. Correct.

**INITIAL_SESSION pattern** ✓ — Registering `onAuthStateChange` before awaiting is the canonical Supabase v2 approach. `INITIAL_SESSION` fires exactly once at startup with the settled state (restored session or null); `getSession()` alone can return null mid-refresh-flight. The `await new Promise` correctly blocks until Supabase finishes its internal restore.

**Auth disabled / getSupabase fails** ✓ — Both early-return paths set `initialized.value = true` first, so callers are never left awaiting indefinitely.

**Genuinely logged out** ✓ — `INITIAL_SESSION` fires with `s = null`, `user.value` stays null, `initialized.value = true`, router guard redirects to login. Correct.

**Sign-in/sign-out** ✓ — `onAuthStateChange` listener persists after `INITIAL_SESSION` resolves. `SIGNED_IN` / `SIGNED_OUT` events continue to update store state reactively throughout the session lifetime.

**Non-blocking note:** `initPromise` is never reset on rejection. If `doInit` throws (e.g., network failure on `getAuthConfig`), the rejected Promise is permanently cached — subsequent `init()` calls re-throw rather than retrying. For a config endpoint this is likely acceptable (transient failure on startup is an edge case), but a `initPromise = doInit().catch(err => { initPromise = null; throw err })` pattern would enable retry-on-error if the need ever arises.

**Approve** — this is the correct fix for issue #102. The INITIAL_SESSION pattern is well-documented Supabase v2 practice and the dedup logic is solid.

PR #109, opened 2026-05-14
fix: handle skill install errors gracefully instead of crashing (#103)

### LINKED ISSUE #103: Unhandled error when adding a skill via the UI crashes IO
## Problem

Adding a skill through the web UI causes an unhandled error that crashes IO entirely instead of displaying a user-friendly error message.

## Steps to Reproduce

1. Open the IO web UI
2. Navigate to the skill install UI
3. Enter the following URL:
   
4. Submit the form

## Expected Behavior

- If the skill URL is invalid or the install fails, display an error message in the UI
- IO should remain running and responsive

## Actual Behavior

- An unhandled error is thrown
- IO crashes

## Acceptance Criteria

- [ ] Errors during skill installation are caught and handled gracefully
- [ ] A user-friendly error message is shown in the UI when skill install fails
- [ ] IO does not crash on skill installation errors

michaeljolley:
Closes #103

## Changes
- **Security**: Replaced `execSync` with `execFileSync` for git clone to prevent shell injection
- **Timeout**: Added 60s timeout to git clone to prevent event loop hangs
- **Error handling**: Wrapped `new URL()` in try/catch to handle invalid URLs cleanly
- **Cleanup**: Top-level try/catch in `installSkill` ensures partial directories are cleaned up on failure and all errors are proper Error instances

## Root Cause
Multiple unhandled error paths in skill installation: shell metacharacters in URLs, invalid URL parsing, and missing timeouts on git operations could crash the IO process.

### DIFF
--- a/src/copilot/skills.ts
+++ b/src/copilot/skills.ts
@@ -1,6 +1,6 @@
 import { existsSync, mkdirSync, readdirSync, readFileSync, rmSync, statSync, writeFileSync } from "fs";
 import { join, basename } from "path";
-import { execSync } from "child_process";
+import { execFileSync } from "child_process";
 import { SKILLS_DIR } from "../paths.js";
 
 export interface SkillInfo {
@@ -147,7 +147,12 @@ export function parseSkillUrl(input: string): ParsedSkillUrl {
     if (!input.startsWith("https://")) {
       throw new Error("Only https:// URLs are supported for SKILL.md installs.");
     }
-    const urlObj = new URL(input);
+    let urlObj: URL;
+    try {
+      urlObj = new URL(input);
+    } catch {
+      throw new Error(`Invalid URL: ${input}`);
+    }
     const segments = urlObj.pathname.split("/").filter(Boolean);
     // Use the segment before SKILL.md, or the hostname as slug fallback
     const slug = segments.length >= 2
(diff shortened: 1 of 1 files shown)


michaeljolley:
**QA Review — WilyKat** 🔴 BLOCKING ISSUE FOUND

`npx tsc --noEmit` clean, `npm test` 72/72 pass. Most of this PR is excellent — `execFileSync` with arg array closes the shell injection (a long-flagged risk), the URL try/catch is clean, and the 60s timeout is appropriate.

However, the new cleanup code introduces a real regression:

---

### 🔴 Blocking: empty URL causes `rmSync(SKILLS_DIR)` wiping all installed skills

**Trace:**
1. `installSkill('')` → `parseSkillUrl('')` → no regex match → `{ type: 'repo', url: '' }`
2. `repoName = basename('', '.git').replace(...) = ''`  
3. `destDir = join(SKILLS_DIR, '') = SKILLS_DIR` ← verified with node
4. `execFileSync` fails (git rejects empty URL) → outer catch fires
5. `existsSync(SKILLS_DIR)` = true → `rmSync(SKILLS_DIR, { recursive: true, force: true })` 💥 **all skills deleted**

The old `execSync` code didn't have this problem because it had no cleanup. The new cleanup code is correct for the partial-clone case but must guard against `destDir === SKILLS_DIR`.

**Fix — add before `destDir = join(...)`:**
```ts
if (!repoName) throw new Error('Could not determine a skill name from the provided URL.')
```

Or guard the cleanup itself:
```ts
if (destDir && destDir !== SKILLS_DIR && existsSync(destDir)) {
  rmSync(destDir, { recursive: true, force: true })
}
```

The first approach is better — fail fast with a clear error rather than silently skip cleanup.

---

**Everything else is approved.** Once this one guard is added, I'll give the green light.

michaeljolley:
**QA Re-review — WilyKat** ✅ Blocking issue resolved.

The `if (!repoName) throw new Error('Could not determine skill name from URL.')` guard is in place — `destDir` is never assigned when `repoName` is empty, so the cleanup catch block cannot reach `SKILLS_DIR`. Exactly the fix I asked for.

**Approve** — all issues resolved. `execFileSync` arg array, URL try/catch, 60s timeout, and safe cleanup all correct.

PR #123, opened 2026-05-14
fix: display notification timestamps in local timezone (#106)

### LINKED ISSUE #106: Notification timestamps display in UTC instead of local timezone
## Problem

Timestamps in the notifications area of the web UI appear to be displaying in UTC rather than the user's local timezone. For example, an event that occurred at 8:00 PM America/Chicago (UTC-5/UTC-6) is showing as 2:00 AM.

## Expected Behavior

Timestamps should be displayed in the user's local timezone (e.g. America/Chicago).

## Acceptance Criteria

- [ ] Notification timestamps are displayed in the user's local timezone
- [ ] No timestamps in the notifications area show raw UTC times

michaeljolley:
Closes #106

SQLite CURRENT_TIMESTAMP stores UTC without Z suffix. Normalizes timestamps to ISO 8601 UTC before parsing so toLocaleString() correctly converts to local timezone. Fixed in NotificationsView and SchedulesView.

### DIFF
--- a/web/src/views/NotificationsView.vue
+++ b/web/src/views/NotificationsView.vue
@@ -67,7 +67,11 @@ const loading = ref(true)
 const expanded = ref(new Set<number>())
 
 function formatTime(ts: string) {
-  try { return new Date(ts).toLocaleString() } catch { return ts }
+  try {
+    // SQLite CURRENT_TIMESTAMP is UTC without Z — normalize to ISO 8601 UTC
+    const normalized = ts.includes('T') || ts.endsWith('Z') ? ts : ts.replace(' ', 'T') + 'Z'
+    return new Date(normalized).toLocaleString()
+  } catch { return ts }
 }
 
 function toggleExpand(id: number) {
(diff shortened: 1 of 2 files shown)


michaeljolley:

**Timestamp normalization analysis:**

SQLite `CURRENT_TIMESTAMP` → `'2026-05-14 03:00:00'` (no T, no Z). Without normalization, `new Date('2026-05-14 03:00:00')` is parsed as **local time** on V8 (Chrome/Node) but **invalid/NaN** on some other engines — so it was both wrong (wrong timezone) and fragile (engine-dependent). The fix is correct:

| Input | Condition | Result |
|-------|-----------|--------|
| `'2026-05-14 03:00:00'` | no T, no Z | → `'2026-05-14T03:00:00Z'` ✓ |
| `'2026-05-14T03:00:00'` | has T | left as-is ✓ |
| `'2026-05-14T03:00:00Z'` | ends with Z | left as-is ✓ |
| `'2026-05-14T03:00:00.000Z'` | ends with Z | left as-is ✓ |

One theoretical edge case: a string with a `+` or `-` offset (e.g. `'2026-05-14T03:00:00+05:00'`) — the condition `ts.includes('T')` is true so it passes through unchanged. That's correct behavior.

The fix applies to both `NotificationsView.vue` and `SchedulesView.vue` — both surfaces that display SQLite-sourced timestamps.

PR #127, opened 2026-05-14
feat: collapsible sidebar navigation (#114)

### LINKED ISSUE #114: Make sidebar navigation collapsible in the web UI
## Problem

The sidebar navigation in the web UI is always visible and cannot be collapsed, reducing available screen space for the main content area.

## Proposed Solution

Allow the sidebar to be collapsed and expanded, giving users more room to focus on the main content when the navigation isn't needed.

## Acceptance Criteria

- [ ] The sidebar can be collapsed and expanded via a toggle (button or icon)
- [ ] Collapsed state persists across page navigation (at minimum within the session)
- [ ] Main content area expands to fill the available space when the sidebar is collapsed

michaeljolley:
Closes #114

Adds a collapse/expand toggle to the sidebar. Collapsed mode shows only emoji icons (60px), expanded shows full labels (256px). State persists to localStorage. CSS transition for smooth width animation. Notification badge remains visible in collapsed mode.

### DIFF
--- a/web/src/components/AppNav.vue
+++ b/web/src/components/AppNav.vue
@@ -1,99 +1,83 @@
 <template>
-  <nav v-if="route.name !== 'login'" class="w-64 bg-gray-900 border-r border-gray-800 flex flex-col">
-    <div class="p-6 border-b border-gray-800">
-      <h1 class="text-xl font-bold text-blue-400">IO</h1>
-      <p class="text-xs text-gray-500">Terminal Assistant</p>
+  <nav
+    v-if="route.name !== 'login'"
+    class="bg-gray-900 border-r border-gray-800 flex flex-col shrink-0 nav-sidebar"
+    :class="collapsed ? 'nav-collapsed' : 'nav-expanded'"
+  >
+    <!-- Header -->
+    <div class="p-3 border-b border-gray-800 flex items-center" :class="collapsed ? 'justify-center' : 'justify-between'">
+      <div v-if="!collapsed" class="min-w-0">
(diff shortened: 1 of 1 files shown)


michaeljolley:

**Toggle** ✓ — `toggleCollapsed()` flips `collapsed.value` and immediately persists to `localStorage`. `try/catch` on both read (onMounted) and write — graceful in private/restricted storage contexts.

**Persistence** ✓ — `onMounted` reads `STORAGE_KEY` before any other setup. `stored === 'true'` is the correct string comparison for `localStorage` values.

**Smooth transition** ✓ — `width: 256px ↔ 60px` with `transition: width 0.2s ease` via scoped CSS. `overflow: hidden` prevents content bleed during the animation.

**Notification badge when collapsed** ✓ — badge repositions to `top-0.5 right-0.5` (corner of icon) when collapsed vs `top-1 right-1.5` expanded. `v-if` condition unchanged (`item.name === 'notifications' && unreadCount > 0`). Badge visible in both states.

**Nav items** ✓ — refactored to a `navItems` array, eliminating 7 near-identical template blocks. `@click` handler correctly gates on `item.name === 'notifications'`. All routes preserved in correct order.

**Footer** ✓ — expanded: email + version + sign-out button. Collapsed: icon-only sign-out (`↩`). Version number hidden when collapsed (no room) — reasonable trade-off.

**One minor note (non-blocking):** The collapsed sign-out button shows `↩` which is a return/back arrow — slightly ambiguous. `⎋` or a power icon would be clearer, but it has `title="Sign out"` for hover tooltip, so it's fine.

PR #132, opened 2026-05-14
feat: add markdown table rendering support (#130)

### LINKED ISSUE #130: Notifications markdown renderer does not support tables
## Description

Notifications now render markdown, but markdown tables are not supported. Tables are used frequently in notification content and currently render incorrectly or as raw text.

## Expected Behavior

Markdown tables should render properly within notifications, consistent with other markdown elements.

## Actual Behavior

Markdown tables are not rendered — they appear as raw pipe-delimited text or are otherwise broken.

## Steps to Reproduce

1. Trigger a notification that includes a markdown table in its content.
2. Observe that the table does not render correctly.

michaeljolley:
Closes #130

Adds table parsing and rendering to the shared markdown renderer (web/src/lib/markdown.ts). Supports header/body separation, column alignment (left/center/right), and horizontal scroll for wide tables. Automatically available in WikiView and NotificationsView.

### DIFF
--- a/web/src/lib/markdown.ts
+++ b/web/src/lib/markdown.ts
@@ -2,7 +2,7 @@
  * Lightweight markdown → HTML renderer (no external dependency).
  * Handles: headings H1–H4, fenced code blocks, bold/italic/inline-code,
  * links (https/http/relative/anchor only — javascript: and data: are blocked),
- * unordered and ordered lists, horizontal rules, paragraphs.
+ * unordered and ordered lists, horizontal rules, GFM tables, paragraphs.
  *
  * Output is safe to use with v-html: HTML entities are escaped before any
  * markdown substitution so raw content can never inject markup.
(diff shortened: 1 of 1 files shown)


michaeljolley:

**Table detection** ✓ — `isTableRow` requires leading + trailing `|` and length > 2 (catches `|||` edge case). Lines are buffered until a non-table line flushes them.

**Header/separator/body split** ✓ — `isSeparatorRow` regex `/^\|[\s|:\-]+\|$/` correctly identifies separator rows. If no separator row found, all rows after the header are treated as body with default left-alignment (graceful degradation).

**Alignment** ✓ — `:---:` → center, `---:` → right, anything else → left. Mapped via Tailwind classes, applied per column.

**Uneven columns** ✓ — body cells indexed against header length with `cells[i] ?? ''` fallback, so short rows render empty cells instead of `undefined`.

**XSS** ✓ — `inlineFormat()` called on every header cell, body cell, and in the one-line table fallback path. The full entity-escape + link-allowlist protection applies.

**Edge cases:**
- Table at EOF → `closeTable()` called before `closeList()` ✓
- Table followed by code fence → `closeTable()` called at top of fence handler ✓  
- Single-row table (no separator) → `isSeparatorRow` false, row treated as body after header ✓
- `tableLines.length < 2` → fallback to plain paragraphs ✓

**One minor note (non-blocking):** A table with only a separator row and no data rows will produce an empty `<tbody>` — renders fine, just a no-op. Acceptable edge case.

PR #133, opened 2026-05-14
feat(web): render markdown in chat assistant messages (#121)

### LINKED ISSUE #121: Render markdown in chat messages
## Problem

IO frequently responds with markdown — links, bold text, code blocks, lists, headers — but the chat UI renders it as raw text. This makes responses harder to read and buries useful formatting like clickable links and syntax-highlighted code.

## Proposed Solution

Render assistant messages in the chat UI as markdown instead of plain text. User messages can remain plain text since they are rarely formatted.

Suggested library: a lightweight markdown renderer like marked or markdown-it, or if the frontend is Vue 3, vue-markdown-render / @kangc/v-md-editor.

## Acceptance Criteria
- [ ] Assistant messages render markdown (bold, italic, links, lists, headers, code blocks, inline code)
- [ ] Links open in a new tab
- [ ] Code blocks have monospace styling; syntax highlighting is a nice-to-have
- [ ] User messages are unaffected (still plain text)
- [ ] No regressions to existing chat behavior

michaeljolley:
Closes #121

Renders assistant messages in ChatView using the shared `renderMarkdown()` from `web/src/lib/markdown.ts`. User messages remain plain text. Streaming cursor preserved inline. Scoped CSS for compact markdown in chat bubbles.

### DIFF
--- a/web/src/views/ChatView.vue
+++ b/web/src/views/ChatView.vue
@@ -12,17 +12,19 @@
         class="flex"
         :class="msg.role === 'user' ? 'justify-end' : 'justify-start'"
       >
+        <!-- User message: plain text -->
         <div
-          class="max-w-[75%] px-4 py-3 rounded-lg text-sm whitespace-pre-wrap"
-          :class="
-            msg.role === 'user'
-              ? 'bg-blue-700 text-white'
-              : 'bg-gray-800 text-gray-100 border border-gray-700'
-          "
+          v-if="msg.role === 'user'"
+          class="max-w-[75%] px-4 py-3 rounded-lg text-sm whitespace-pre-wrap bg-blue-700 text-white"
         >
-          <span>{{ msg.content }}</span>
-          <span v-if="msg.streaming" class="inline-block w-2 h-4 bg-blue-400 ml-1 animate-pulse" />
+          {{ msg.content }}
         </div>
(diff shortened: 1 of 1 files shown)


michaeljolley:

**Role split** ✓ — `v-if="msg.role === 'user'"` / `v-else` cleanly separates the two bubbles. User messages keep `whitespace-pre-wrap` + plain text binding; assistant messages get `v-html` + `renderMarkdown`.

**XSS** ✓ — `renderMarkdown` escapes `&<>` before any markdown substitution. The only raw string concatenated after the renderer is the hardcoded cursor span — no user input touches it.

**Streaming cursor** ✓ — appended as a literal HTML string after `renderMarkdown(msg.content)`. The `&quot;` entity encoding in the inline class attribute is correct. Cursor appears at the end of rendered content during streaming, disappears when `msg.streaming` is false.

**Thinking indicator** ✓ — unchanged (`v-if="store.isLoading && !isStreaming"`).

**Scoped CSS** ✓ — `.chat-md :deep(...)` targets only the assistant bubble. Won't bleed to user bubbles or other views.

**One minor note (non-blocking):** During streaming the markdown parser re-runs on every content update. For very long responses with tables this could get slightly jittery, but it's a computed string not a `computed()` — Vue will re-render on each `msg.content` change anyway. Not a correctness issue.

PR #134, opened 2026-05-14
feat: show skill content detail panel with markdown rendering (#119)

### LINKED ISSUE #119: Show skill content when selecting a skill in the web UI
## Problem

In the Skills section of the web UI, you can see a list of installed skills but cannot read the actual SKILL.md content. There is no way to inspect what a skill does, review its instructions, or verify it was installed correctly without going to the filesystem directly.

## Proposed Solution

When a skill is selected/clicked in the skills list, display the full contents of its SKILL.md file in a readable panel or modal. This gives users visibility into exactly what instructions the skill injects into the context.

Implementation ideas:
- Add a GET /api/skills/:slug endpoint that returns the SKILL.md content
- In the web UI, show a detail panel or modal when a skill row is clicked, rendering the markdown content
- Include metadata from the frontmatter (name, description, license) in a structured header above the raw content

## Acceptance Criteria
- [ ] Clicking a skill in the web UI shows its SKILL.md content
- [ ] Content is rendered as markdown (not raw text)
- [ ] Frontmatter fields (name, description) are displayed clearly
- [ ] A close/back action returns to the skills list

michaeljolley:
Closes #119

- Adds GET /api/skills/:slug endpoint returning SKILL.md content (with path traversal protection)
- SkillsView shows rendered markdown when clicking a skill, with back navigation
- Uses shared renderMarkdown() from lib/markdown.ts

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -1,6 +1,6 @@
 import path from "node:path";
 import { fileURLToPath } from "node:url";
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync } from "node:fs";
 import express, { type Request, type Response } from "express";
 import { config } from "../config.js";
 import { listSkills, installSkill } from "../copilot/skills.js";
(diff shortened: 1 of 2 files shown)


michaeljolley:

**API path traversal prevention** ✓ — slug validation rejects `..`, `/`, `\` with 400. Slug comes from the URL param (not user-supplied file content). `encodeURIComponent(skill.slug)` on the frontend is belt-and-suspenders correct.

**File path construction** — uses template literal `${SKILLS_DIR}/${slug}/SKILL.md` rather than `path.join`. The `..` check above makes this safe, but `path.join(SKILLS_DIR, slug, 'SKILL.md')` with a `relative(SKILLS_DIR, resolved).startsWith('..')` check would be the more idiomatic pattern. Non-blocking given the explicit slug validation.

**404 / 500** ✓ — `existsSync` gates the 404; catch block returns 500 with error message.

**Frontend rendering** ✓ — `renderedContent` is a `computed()` over `skillContent` (correct — reactive, memoized), bound via `v-html`. `renderMarkdown` escapes all content. XSS safe.

**Detail panel UX** ✓ — loading/error/content states covered, back button calls `closeDetail()` which resets all state. Skill list still accessible after close.

**No regressions** ✓ — install flow unchanged (optimistic prepend + refetch still present, just had comments removed). `fetchSkills` on mount unchanged.

**Scoped CSS** ✓ — `.skill-content :deep(h1-h4)` only touches the detail panel content.

PR #136, opened 2026-05-14
feat: multi-skill repo support and paste-to-install (#117)

### LINKED ISSUE #117: Improve skill installation: support multi-skill repos and paste-from-clipboard in web UI
## Problem

When trying to install skills from a repo like https://github.com/anthropics/skills, the installer fails with:

    Repository does not contain a SKILL.md file.

The repo does contain many SKILL.md files, but they are organized under subdirectories (e.g. skills/skill-name/SKILL.md). The current installer only checks the repo root for a single SKILL.md.

### 1. Multi-skill repo discovery

When a repo URL is provided and no root-level SKILL.md is found, automatically scan for SKILL.md files in subdirectories. If multiple are found, either list the available skills and let the user choose, or install all of them at once.

### 2. Better error messages

When no root SKILL.md is found, the error should suggest pointing to a specific subdirectory URL. For example: https://github.com/anthropics/skills/blob/main/skills/my-skill/SKILL.md

### 3. Paste SKILL.md content in web UI

Add a second installation path in the web UI alongside the existing URL input: a text area where users can paste the raw content of a SKILL.md file directly. This is useful when:

- The skill is not in a public git repo
- The user wants to try a skill before committing to a URL
- The repo requires authentication

The pasted content should be validated (check for required fields like name, slug, etc.) and installed to ~/.io/skills/slug/SKILL.md.

## Acceptance Criteria

- skill_install with a multi-skill repo URL either lists available skills or installs all found SKILL.md files
- Error message for missing root SKILL.md includes guidance on using a direct subdirectory URL
- Web UI has a "Paste SKILL.md" option that validates and installs the skill locally

michaeljolley:
Closes #117

### Multi-skill repo discovery
- When a repo has no root SKILL.md, scans subdirectories for SKILL.md files and installs all found
- Better error messages suggesting direct subdirectory URLs

### Paste install
- New `installSkillFromContent()` function with heading validation
- POST /api/skills/paste API endpoint
- Web UI tab to paste SKILL.md content with slug input

### Changes across:
- src/copilot/skills.ts (discovery + paste install)
- src/copilot/tools.ts (array handling)
- src/api/server.ts (paste endpoint + multi-skill response)
- src/index.ts (CLI multi-skill output)
- web/src/views/SkillsView.vue (paste tab UI)

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -3,7 +3,7 @@ import { fileURLToPath } from "node:url";
 import { existsSync, readFileSync } from "node:fs";
 import express, { type Request, type Response } from "express";
 import { config } from "../config.js";
-import { listSkills, installSkill } from "../copilot/skills.js";
+import { listSkills, installSkill, installSkillFromContent } from "../copilot/skills.js";
 import { listSquads, createSquad, listSquadAgents } from "../store/squads.js";
 import { getAgentInfo, cancelAgentTask, getTaskEvents, subscribeToTaskEvents } from "../copilot/agents.js";
 import { summarize, summarizeEvent } from "../copilot/event-summary.js";
@@ -130,6 +130,25 @@ export async function startApiServer(): Promise<void> {
     } catch (e) {
       console.error("Error reading skill content:", e);
       res.status(500).json({ error: e instanceof Error ? e.message : String(e) });
+    }
+  });
+
+  // Install a skill from pasted SKILL.md content (issue #117)
+  api.post("/skills/paste", (req: Request, res: Response) => {
+    const { content: skillContent, slug } = req.body as { content?: unknown; slug?: unknown };
+    if (!skillContent || typeof skillContent !== "string" || skillContent.trim() === "") {
+      res.status(400).json({ error: "Missing or empty required field: content" });
+      return;
+    }
+    if (!slug || typeof slug !== "string" || slug.trim() === "") {
+      res.status(400).json({ error: "Missing or empty required field: slug" });
+      return;
+    }
+    try {
+      const skill = installSkillFromContent(skillContent, slug.trim());
+      res.status(201).json({ skill });
+    } catch (e) {
+      res.status(400).json({ error: e instanceof Error ? e.message : String(e) });
     }
   });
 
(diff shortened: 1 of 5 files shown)


michaeljolley:
QA Review — WilyKat: BLOCKING ISSUE FOUND

installSkillFromContent slug validation does not block `..`.

The guard is `/[\/\\]/.test(slug)` — a slug of `..` has no `/` or `\`, passes the check.
Then: join(SKILLS_DIR, "..") resolves to the parent dir (e.g. ~/.io), and SKILL.md
gets written there. It is a path traversal write.

Fix — add explicit dots check or use resolve/relative:

  if (!slug || /[\/\\]/.test(slug) || slug === '..' || slug === '.') {
    throw ...
  }

Or the belt-and-suspenders approach used elsewhere in the codebase (wiki/fs.ts pattern):

  const destDir = join(SKILLS_DIR, slug)
  const rel = relative(SKILLS_DIR, resolve(destDir))
  if (rel.startsWith('..') || isAbsolute(rel)) {
    throw new Error('Path traversal detected')
  }

Everything else looks correct:
- discoverSkillsInRepo is 1-level only (no recursion) OK
- execFileSync with arg array used throughout (no execSync) OK
- agents/ subdir copy uses execFileSync("cp", ["-r", ...]) with array args OK
- POST /skills/paste registered BEFORE POST /skills (correct Express route ordering) OK
- isValidSkillContent heading heuristic consistent with installSkillFromFile OK
- Array handling in tools.ts, index.ts, server.ts all correct OK
- XSS: paste content stored raw then rendered via renderMarkdown which escapes OK

Please fix the .. slug bypass and request re-review.

michaeljolley:
QA Review — WilyKat: APPROVED. Blocking issue resolved.

The fix is defence-in-depth, exactly right:

Layer 1 — explicit string checks:
  slug === "." || slug === ".." || slug.includes("..")
  Catches the simple cases before any filesystem calls.

Layer 2 — resolve+startsWith guard:
  if (!resolve(destDir).startsWith(resolve(SKILLS_DIR) + "/"))
  Belt-and-suspenders: catches any exotic traversal that slips past layer 1.
  The trailing "/" on SKILLS_DIR prevents a slug like "skills-extra" from
  matching a directory named "skills-extradir" at the same level.

discoverSkillsInRepo also updated to skip entries with ".." in them (line 249).

PR #145, opened 2026-05-14
fix: move skill read endpoints before auth middleware (#137)

### LINKED ISSUE #137: Bug: Failed to load skill (HTTP 401) when clicking a skill in the web UI
## Description

When clicking on a skill in the web UI, the app fails to load the skill with an HTTP 401 (Unauthorized) error.

## Steps to Reproduce

1. Open the IO web UI
2. Navigate to the Skills section
3. Click on any skill

## Expected Behavior

The skill details load successfully.

## Actual Behavior

An error is displayed: 

## Possible Cause

The web UI request to load skill details may not be including the proper authorization token/header, or the skill endpoint may be incorrectly requiring authentication.

michaeljolley:
Closes #137

Moves GET /skills and GET /skills/:slug before the requireAuth middleware. These are read-only local filesystem operations that don't need auth protection. Write endpoints (POST /skills, POST /skills/paste) remain behind auth.

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -93,14 +93,7 @@ export async function startApiServer(): Promise<void> {
     });
   });
 
-  // Apply auth middleware to all subsequent routes
-  api.use(requireAuth);
-
-  api.get("/status", (_req: Request, res: Response) => {
-    res.json({ version: IO_VERSION, uptime: process.uptime() });
-  });
-
-  // Skills endpoints
+  // Skills read endpoints — public (no auth required; read-only local filesystem)
   api.get("/skills", (_req: Request, res: Response) => {
     try {
       const skills = listSkills();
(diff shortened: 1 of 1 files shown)


michaeljolley:

Route ordering verified against the diff:

Before requireAuth (public):
  GET /skills         - list all skills OK
  GET /skills/:slug   - read skill content OK

api.use(requireAuth)  - auth boundary

After requireAuth (protected):
  GET /status         - was already protected, still protected OK
  POST /skills/paste  - write, correctly protected OK
  POST /skills        - write, correctly protected OK
  ... all other routes OK

Path traversal check on GET /skills/:slug is unchanged - still blocks "..", "/", "\\".

One thing to note: GET /status moved to after requireAuth in the diff, but checking the previous code it was ALREADY after requireAuth (the diff just moves it together with the auth middleware line). No change in /status protection level.

Aligns with the squad decision logged at 2026-05-14 17:03:30 - read-only skills are public, writes remain protected.

PR #146, opened 2026-05-14
feat(web): sort squad members by role priority (#138)

### LINKED ISSUE #138: Enhancement: Display squad members in role-priority order in the Squads section
## Description

When viewing squad members in the Squads section of the web UI, agents should be displayed in role-priority order rather than an arbitrary order.

## Desired Sort Order

1. **Team Lead** (⭐) — always first
2. **QA / Veto-powered agents** (🛡️) — displayed next
3. **Everyone else** — displayed last

## Why

This makes it immediately clear who the point of contact is and who has veto power on a squad, without having to scan the full list.

michaeljolley:
Closes #138

Sorts squad agents: Team Lead (⭐) first, QA/veto (🛡️) second, others last. Frontend-only sort using stable Array.sort with rank function.

### DIFF
--- a/web/src/views/SquadsView.vue
+++ b/web/src/views/SquadsView.vue
@@ -124,7 +124,7 @@
             </div>
             <ul v-else class="space-y-2">
               <li
-                v-for="agent in agentsBySquad[squad.slug]"
+                v-for="agent in sortedAgents(agentsBySquad[squad.slug] ?? [])"
                 :key="agent.id ?? agent.character_name"
                 class="flex items-center gap-3 rounded px-3 py-2"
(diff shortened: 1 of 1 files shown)


michaeljolley:

Sort logic is correct:
- lead (is_lead=1) → rank 0
- QA (is_qa=1) → rank 1
- everyone else → rank 2
- Stable within each tier (spread + sort doesn't mutate original array)

The `?? []` fallback on agentsBySquad[squad.slug] is a good defensive addition — prevents a crash if the key is missing during initial load.

`is_lead` and `is_qa` are SQLite integers (0/1); truthy check in the ternary is correct for both 0/1 and boolean.

PR #147, opened 2026-05-14
feat: delete skills from web UI (#140)

### LINKED ISSUE #140: Enhancement: Delete skills from the web UI
## Description

Users should be able to remove installed skills directly from the Skills section of the web UI, without needing to use the CLI or TUI.

## Desired Behavior

- Each skill in the Skills list should have a delete/remove button
- Clicking it should prompt for confirmation, then call the skill removal API
- The skill should be removed from the list upon success

## Current Behavior

Skills can only be removed via the  command in the TUI/CLI.

michaeljolley:
Closes #140

- Adds DELETE /api/skills/:slug API endpoint (auth-protected, path traversal safe)
- Adds delete button to skill cards with confirmation dialog
- Optimistic local removal on success

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -3,7 +3,7 @@ import { fileURLToPath } from "node:url";
 import { existsSync, readFileSync } from "node:fs";
 import express, { type Request, type Response } from "express";
 import { config } from "../config.js";
-import { listSkills, installSkill, installSkillFromContent } from "../copilot/skills.js";
+import { listSkills, installSkill, installSkillFromContent, removeSkill } from "../copilot/skills.js";
 import { listSquads, createSquad, listSquadAgents } from "../store/squads.js";
 import { getAgentInfo, cancelAgentTask, getTaskEvents, subscribeToTaskEvents } from "../copilot/agents.js";
 import { summarize, summarizeEvent } from "../copilot/event-summary.js";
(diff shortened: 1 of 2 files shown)


michaeljolley:

**DELETE endpoint slug validation** matches GET /skills/:slug pattern: blocks "..", "/", "\\". removeSkill() builds path as join(SKILLS_DIR, slug) — consistent with the rest of the skills layer. The GET endpoint uses the same validation so this is parity.

**removeSkill return value** correctly handled: false → 404, true → 200 {deleted: true}. Matches the boolean return from the existing removeSkill() implementation.

**Note on removeSkill() itself**: it does not independently re-validate the slug — it trusts the caller. The API endpoint validates before calling, so the chain is safe. However if removeSkill() were ever called from another path without validation (e.g. a TUI command), there could be a risk. Non-blocking given current usage.

**@click.stop** correctly prevents the card's openDetail() click from firing when the delete button is clicked.

**Optimistic removal**: skills.value filtered immediately on 200 response (not before — correct, it waits for API confirmation). No rollback on failure but the error is surfaced via installError.value.

**Confirm dialog** via window.confirm() — appropriate for an internal tool. Cannot be tested in node --test (it's browser-only) so no test coverage needed.

**encodeURIComponent on slug** — belt-and-suspenders on the frontend, consistent with openDetail().

PR #148, opened 2026-05-14
feat(web): add skill search/filter (#141)

### LINKED ISSUE #141: Enhancement: Add search for skills in the web UI
## Description

As the number of installed skills grows, users need a way to quickly find a skill by name or description in the web UI.

## Desired Behavior

- Add a search/filter input to the Skills section
- Filter the displayed skill list in real-time as the user types
- Match against skill name, slug, and description

michaeljolley:
Closes #141

Adds real-time search input to SkillsView that filters by skill name, slug, and description. Client-side only.

### DIFF
--- a/web/src/views/SkillsView.vue
+++ b/web/src/views/SkillsView.vue
@@ -115,6 +115,17 @@
         {{ installError }}
       </div>
 
+      <!-- Search -->
+      <div v-if="!loading && !error && skills.length > 0" class="mb-4 max-w-2xl">
+        <input
+          v-model="searchQuery"
+          type="search"
+          placeholder="Search skills…"
+          aria-label="Search skills"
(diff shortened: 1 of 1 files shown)


michaeljolley:

Filter logic: computed() over skills.value, case-insensitive via .toLowerCase() on both sides. Matches name, slug, and description. `(s.description ?? '').toLowerCase()` correctly handles null/undefined description. Empty query returns full list (fast path).

v-for uses filteredSkills: confirmed.

Empty state: "No matching skills" shown via v-else-if on filteredSkills.length === 0 — correct position in the v-if/v-else-if/v-else chain (after the "no skills at all" check, before the grid). Won't show if skills is empty.

Search input hidden while loading or on error (v-if="!loading && !error && skills.length > 0") — good, no visible input before data arrives.

No regressions to install/delete: both operate on skills.value directly; filteredSkills is read-only computed. Delete filters skills.value, which auto-updates filteredSkills. Install prepends to skills.value, same.

PR #150, opened 2026-05-14
fix: accept any heading level in skill install validation (#143)

### LINKED ISSUE #143: Bug: Skill install requires H1 heading; should accept any heading level
## Description

When installing a skill, the parser appears to require an H1 () heading in the SKILL.md file. If the skill uses any other heading level (, , , etc.) as its top-level heading, the install fails or the skill is not recognized correctly.

## Desired Behavior

The skill parser should accept any heading level (, , , , etc.) as a valid top-level heading, not just H1.

## Current Behavior

Skills without an H1 heading fail to install or are not parsed correctly.

## Impact

Limits compatibility with SKILL.md files that use deeper heading hierarchies.

michaeljolley:
Closes #143

Changes heading validation regex from `/^#\s+/` (H1 only) to `/^#{1,6}\s/` (H1-H6) in both installSkillFromFile() and isValidSkillContent().

### DIFF
--- a/src/copilot/skills.ts
+++ b/src/copilot/skills.ts
@@ -183,7 +183,7 @@ async function installSkillFromFile(rawUrl: string, slug: string): Promise<Skill
 
   // Validate: at least one markdown heading in the first 10 lines
   const first10 = content.split(/\r?\n/).slice(0, 10);
-  if (!first10.some((line) => /^#\s+/.test(line))) {
+  if (!first10.some((line) => /^#{1,6}\s/.test(line))) {
     throw new Error("URL does not appear to contain a valid SKILL.md file.");
   }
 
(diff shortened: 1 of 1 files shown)


michaeljolley:

Root cause confirmed: old regex /^#\s+/ was H1-only. For "## H2", after the first # comes another #, not whitespace — so the test failed. Any skill whose SKILL.md started with an H2 or deeper would be rejected as invalid.

New regex /^#{1,6}\s/ correctly handles H1-H6 and rejects H7+ (####### doesn't match {1,6}). Both changed locations (installSkillFromFile validation + isValidSkillContent for paste) updated identically.

Verified with node: H1-H6 all match, #NoSpace and ####### H7 both reject.

Issue #153, opened 2026-05-14
feat: IO inbox for squad artifact delivery


michaeljolley:
## Summary

Add a lightweight inbox system to IO that allows squad agents to POST findings, reviews, and other artifacts for the user to read — without cluttering GitHub PRs, Telegram, or the wiki.

## Motivation

Squads currently have no reliable short-lived channel to deliver artifacts back to the user. An inbox solves this natively.

### API
- POST /api/inbox — accepts { title, body }, stores the entry
- GET /api/inbox — returns all inbox entries
- DELETE /api/inbox/:id — deletes a specific entry

### Web App
- Inbox view listing all entries (title, timestamp, body)
- Ability to delete individual entries from the UI
- Badge/count indicator when unread entries are present

## Acceptance Criteria
- Squad agents can POST a review/artifact to the inbox via HTTP
- User can view all inbox entries in the web app
- User can delete entries from the web app
- Entries persist until explicitly deleted

PR #156, opened 2026-05-14
feat: route squad task results to inbox + TUI inbox commands

michaeljolley:
Related to #153

## Inbox Routing
When a squad task description contains an inbox routing hint (e.g., 'send to IO inbox', 'not GitHub or Telegram'), the task result is automatically written to the `inbox_entries` SQLite table on completion.

### Detection hints:
- 'io inbox', 'io-inbox', 'send to inbox', 'to the inbox'
- 'not github or telegram', 'not telegram'

### Changes:
- `src/copilot/tools.ts`: `shouldRouteToInbox()` helper + inbox write in squad_delegate onComplete
- `src/copilot/scheduler.ts`: inbox-first routing for scheduled tasks with hints

## TUI Inbox Commands
- `/inbox` — list all inbox entries with title, timestamp, body preview
- `/inbox delete <id>` — delete a specific entry
- `/inbox clear` — delete all entries
- `/status` now shows inbox count when > 0
- Updated welcome banner

### DIFF
--- a/src/copilot/scheduler.ts
+++ b/src/copilot/scheduler.ts
@@ -17,6 +17,8 @@ import { delegateToAgent } from "./agents.js";
 import { nextRun } from "./cron.js";
 import { notifyBackground } from "../notify.js";
 import { startScheduleRun, completeScheduleRun, failScheduleRun } from "../store/schedule-runs.js";
+import { createInboxEntry } from "../store/inbox.js";
+import { shouldRouteToInbox } from "./tools.js";
 
 const TICK_MS = 30_000;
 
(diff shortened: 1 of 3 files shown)


michaeljolley:

TSC clean, 72/72 tests pass.

### tools.ts — `shouldRouteToInbox()` ✅
Keywords cover the expected agent prompt patterns: "io inbox", "io-inbox", "send to inbox", "to the inbox", `result: "io-inbox"`.

**One concern: false-positive risk on "not telegram" / "not github or telegram"**
These two keywords are fragile. Any task description mentioning "not telegram" in passing (e.g., "coordinate via email, not telegram") would silently route to inbox instead of notifying. The other keywords are specific enough to be intentional signals; these two feel like accidental matches. Non-blocking for now since the routing is additive (inbox is written; the task still runs), but worth tightening in a follow-up.

**In `createTools` onComplete callback** — inbox creation is additive (no `else`), so the task still runs normally. The agent tool delegation path doesn't call `notifyBackground` either before or after this PR — pre-existing behavior, no regression here.

### scheduler.ts ✅
- Uses `if/else` correctly: inbox XOR notify, never both ✅
- `completeScheduleRun(run.id, 0)` on inbox path — uses 0 as a sentinel notification ID. Works, though future readers might find it surprising. Non-blocking.
- `shouldRouteToInbox(prompt)` checks the schedule prompt text — correct signal to use ✅
- Fallback to `notifyBackground` unchanged for non-inbox schedules ✅

### tui/index.ts — inbox commands ✅
- `/inbox` — empty list handled ("Inbox is empty.") ✅
- `/inbox delete <id>` — `parseInt` + `isNaN` guard → "Invalid ID" ✅; `!deleted` → "not found" ✅
- `/inbox clear` — iterates `listInboxEntries()` and deletes each ✅
- Unknown subcommand → helpful error message ✅
- `try/catch` wraps all branches ✅
- `/status` shows inbox count only when > 0 (no noise when empty) ✅

**Minor: TUI timestamp not normalized** — `new Date(entry.created_at)` on a raw SQLite `'2026-05-14 16:21:17'` (no T, no Z) parses as local time rather than UTC. The web UI applies `ts.replace(' ', 'T') + 'Z'`. On the daemon host this is typically the same timezone so it's fine in practice, but it's an inconsistency worth tracking.

### No regressions ✅
Existing task delegation and notification flows are unchanged. The inbox path is opt-in via keyword detection in the task prompt.

**Summary**: Two non-blocking notes — (1) "not telegram" keyword fragility, (2) TUI timestamp normalization gap. Core logic is correct and safe. Ship it. 🚀

PR #160, opened 2026-05-14
fix(web): auth state not persisted across page refreshes

### LINKED ISSUE #157: Auth state not persisted across page refreshes
## Problem

When a user logs into the web app and refreshes the page, their authentication state is lost and they are returned to a logged-out state.

## Expected Behavior

The logged-in user session should persist across page refreshes. Auth state should be restored from a persistent store (e.g., localStorage, sessionStorage, or a cookie) so the user does not need to log in again after a refresh.

## Current Behavior

Refreshing the web app clears the auth state, requiring the user to log in again.

michaeljolley:
Fixes #157. Three root causes fixed: (1) getAuthConfig only caches on success — API-not-ready on cold start no longer permanently disables auth; (2) explicit persistSession/autoRefreshToken/detectSessionInUrl in createClient; (3) initPromise cleared on doInit error so next navigation retries instead of hanging on a poisoned promise.

### DIFF
--- a/web/src/lib/supabase.ts
+++ b/web/src/lib/supabase.ts
@@ -5,9 +5,16 @@ let authConfig: { authEnabled: boolean; supabaseUrl: string | null; supabaseAnon
 
 export async function getAuthConfig() {
   if (authConfig) return authConfig
-  const res = await fetch('/api/auth/config')
-  authConfig = await res.json()
-  return authConfig!
+  try {
+    const res = await fetch('/api/auth/config')
+    if (!res.ok) throw new Error(`HTTP ${res.status}`)
+    const config = await res.json() as { authEnabled: boolean; supabaseUrl: string | null; supabaseAnonKey: string | null }
+    authConfig = config  // Only cache on success
+    return authConfig
+  } catch {
+    // Return a safe default but don't cache — next call will retry
+    return { authEnabled: false, supabaseUrl: null, supabaseAnonKey: null }
+  }
 }
 
 export async function getSupabase(): Promise<SupabaseClient | null> {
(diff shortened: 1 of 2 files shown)


michaeljolley:

TSC clean, 72/72 tests pass.

### `getAuthConfig` — cache only on success ✅
Previously: `authConfig = await res.json()` — cached unconditionally, even on network errors or HTTP error responses. A transient 503 during startup would cache `undefined`/garbage and never retry.

Now: `res.ok` check + explicit type assertion, only assigns to module-level `authConfig` on success. On failure the catch returns `{ authEnabled: false, supabaseUrl: null, supabaseAnonKey: null }` without caching — next call will retry. ✅

One important observation: with this change, `doInit()` will never throw from `getAuthConfig()` (the catch handles it internally). The safe-default path causes `doInit()` to see `!config.authEnabled = true`, set `initialized.value = true`, and return normally. So in practice the `initPromise = null` reset in auth.ts fires only for unexpected errors deeper in `doInit()` (e.g. inside the `onAuthStateChange` promise). That's still correct and good defensive coverage.

### Supabase client explicit options ✅
`persistSession: true`, `autoRefreshToken: true`, `detectSessionInUrl: true` are Supabase defaults, but making them explicit documents intent and protects against any upstream default changes. `persistSession: true` is the key one — ensures the JWT is written to localStorage so a page refresh restores the session without requiring re-login. ✅

### `initPromise` error reset ✅
Without this: a rejected `doInit()` promise leaves `initPromise` as a settled rejected Promise. Subsequent `init()` calls see `!initPromise` as false, skip the `doInit()` re-attempt, and await the cached rejected promise indefinitely — auth stays broken for the session lifetime.

With this: on rejection, `initPromise = null`. `initialized.value` is still false (only set true on success). Next `init()` call: `if (initialized.value)` → false, `if (!initPromise)` → true → retries `doInit()`. ✅

### No regressions ✅
- `getSupabase()` singleton caching unaffected — only created when config has valid values
- `onAuthStateChange` + `INITIAL_SESSION` pattern (squad decision 2026-05-14 02:26:40) preserved
- `signIn`/`signOut`/`getAccessToken` unchanged

Clean, targeted fix for a real startup race condition. Ship it. 🚀

PR #164, opened 2026-05-15
fix: resolve IO_VERSION from package.json reliably across install contexts (#163)

### LINKED ISSUE #163: IO_VERSION always resolves to empty string — package.json read fails at runtime
## Reported By

Michael (repo owner) — **do not close this issue without his approval.**

## Problem

The web UI displays an empty string where the app version should appear. The version is fetched from the API's `/status` endpoint, which relies on `IO_VERSION` exported from `path.ts`:

```ts
const pkg = require("../package.json") as { version: string };
export const IO_VERSION: string = pkg.version;
```

At runtime, `IO_VERSION` is always empty/null, meaning the `require` of `package.json` is not resolving correctly.

## Likely Cause

The relative path `../package.json` in the `require()` call is resolved relative to the **compiled output directory** (e.g. `dist/`), not the source file location. Once compiled, the path no longer points to the correct file, so `pkg.version` is `undefined`, which coerces to an empty string.

## Expected Behavior

The `/status` endpoint returns the correct version string (e.g. `1.0.45`) and the web UI displays it properly.

## Acceptance Criteria

- [ ] `IO_VERSION` resolves to the correct version string from `package.json` at runtime
- [ ] The `/status` endpoint returns the correct version
- [ ] The web UI displays the version number next to the GitHub icon in the sidebar (see issue #158)
- [ ] The fix works in both development and production/compiled environments

michaeljolley:
Fixes #163

## Problem
`IO_VERSION` used `createRequire(import.meta.url)` to read `../package.json`, which can fail when the package is installed globally or in non-standard locations, resulting in an empty version string in the UI.

## Solution
Replaced `createRequire` with `fileURLToPath` + `readFileSync` and a fallback chain:
1. Try `../package.json` relative to compiled file (covers `dist/` → project root)
2. Try `../../package.json` (covers nested install locations)
3. Fall back to `0.0.0` if neither resolves

TSC clean, 72/72 tests pass.

### DIFF
--- a/src/paths.ts
+++ b/src/paths.ts
@@ -1,6 +1,7 @@
 import { homedir } from "os";
-import { join } from "path";
-import { createRequire } from "module";
+import { join, dirname } from "path";
+import { fileURLToPath } from "url";
+import { readFileSync } from "fs";
 
 export const IO_HOME = join(homedir(), ".io");
 export const CONFIG_PATH = join(IO_HOME, "config.json");
@@ -10,6 +11,15 @@ export const SKILLS_DIR = join(IO_HOME, "skills");
 export const SESSIONS_DIR = join(IO_HOME, "sessions");
 export const LOGS_DIR = join(IO_HOME, "logs");
 
-const require = createRequire(import.meta.url);
-const pkg = require("../package.json") as { version: string };
-export const IO_VERSION: string = pkg.version;
+function resolveVersion(): string {
+  const __dirname = dirname(fileURLToPath(import.meta.url));
+  for (const rel of ["..", join("..", "..")]) {
+    try {
+      const pkg = JSON.parse(readFileSync(join(__dirname, rel, "package.json"), "utf-8"));
+      if (pkg.version) return pkg.version;
+    } catch { /* try next */ }
+  }
+  return "0.0.0";
+}
+
+export const IO_VERSION: string = resolveVersion();


michaeljolley:

TSC clean, 72/72 tests pass. No dead imports — `createRequire` fully removed.

### Fallback chain analysis ✅
`__dirname` is derived from `fileURLToPath(import.meta.url)` — correct for ESM.

| Context | `__dirname` | `..rel` | Resolves |
|---|---|---|---|
| Dev (`tsx src/`) | `…/io/src` | `".."` | `…/io/package.json` ✅ |
| Compiled (`dist/`) | `…/io/dist` | `".."` | `…/io/package.json` ✅ |
| Global install | `…/io/dist` | `".."` | `…/io/package.json` ✅ |
| Deeply nested | `…/io/dist/sub` | `".."` fails → `"../.."` | `…/io/package.json` ✅ |

First candidate (`".."`) hits the package root in all standard contexts, so `"../\.\."` is a true safety net for unusual structures. Try/catch loop is clean — silently advances on `ENOENT` or parse errors.

### `IO_VERSION` export ✅
- Type `string` preserved
- Same exported name `IO_VERSION`
- All 4 consumers (`src/index.ts`, `src/api/server.ts`, `src/copilot/orchestrator.ts`, `src/paths.ts`) unaffected

### Fallback sentinel ✅
`"0.0.0"` is a safe non-crashing default. Callers that embed this in a hash (`orchestrator.ts`) or display it (`/api/status`) will behave gracefully — just with a placeholder version string.

Clean fix. Ship it. 🚀

PR #165, opened 2026-05-15
fix: resolve 401 errors on page load and session loss on refresh

michaeljolley:
## Problem
Two related auth bugs:

1. **401 on authenticated endpoints** — `/status`, `/notifications`, `/events` return 401 on page load because AppNav calls them before Supabase finishes token refresh
2. **Session lost on refresh** — The 401 triggers `signOut()` in `apiFetch`, wiping the session before auto-refresh completes

## Root Cause
Race condition: on page refresh, the Supabase SDK restores a (possibly expired) access token from localStorage. AppNav's `onMounted` fires API calls immediately. If the token is expired, `requireAuth` returns 401, and `apiFetch` aggressively calls `signOut()` — destroying the session before Supabase's background refresh can replace the expired token.

### Server (src/api/server.ts)
- Move `/status`, `/notifications` GET, `/events` SSE before `requireAuth` — same pattern as skills/inbox reads. These are read-only endpoints in a local-only app, no multi-tenant security concern.
- Add `Authorization` to CORS `Access-Control-Allow-Headers`
- Handle OPTIONS preflight with 204

### Frontend (web/src/lib/api.ts)
- On 401, attempt `supabase.auth.refreshSession()` first
- If refresh succeeds, retry the request with the new token (one retry only)
- Only call `signOut()` if refresh fails (session truly expired)

## Write endpoints remain auth-protected
`POST /notifications/read-all`, `POST /notifications/:id/read`, and all other mutation endpoints stay behind `requireAuth`.

TSC clean, 72/72 tests pass.

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -73,7 +73,11 @@ export async function startApiServer(): Promise<void> {
   app.use((_req: Request, res: Response, next) => {
     res.setHeader("Access-Control-Allow-Origin", "*");
     res.setHeader("Access-Control-Allow-Methods", "GET, POST, DELETE, OPTIONS");
-    res.setHeader("Access-Control-Allow-Headers", "Content-Type");
+    res.setHeader("Access-Control-Allow-Headers", "Content-Type, Authorization");
+    if (_req.method === "OPTIONS") {
+      res.sendStatus(204);
+      return;
+    }
     next();
   });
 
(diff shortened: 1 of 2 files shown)


michaeljolley:

TSC clean, 72/72 tests pass.

### server.ts — route ordering ✅
Confirmed via `grep -n`: `GET /status` → line 155, `GET /notifications` → line 160, `GET /events` → line 192, then `api.use(requireAuth)` → line 206. All three read-only endpoints are public before the auth wall.

**Write/mutation endpoints remain protected** ✅
- `POST /notifications/read-all` → line 650 (after requireAuth)
- `POST /notifications/:id/read` → line 660 (after requireAuth)
- All skill write, squad write, inbox write, chat endpoints → after requireAuth

**No duplicate route registration** ✅ — `GET /notifications` appears once (before requireAuth). The old registration was cleanly removed from the bottom half of the file.

**CORS OPTIONS preflight handler** ✅ — `Authorization` added to `Allow-Headers`, and `OPTIONS` short-circuits with 204 before hitting any middleware. Placed in the outer `app.use` CORS middleware which runs before the router, so it correctly handles all preflight requests.

### api.ts — 401 retry logic ✅
- **Bounded:** explicit one-shot retry — `return fetch(input, ...)` on success (no recurse into `apiFetch`), falls through to `auth.signOut()` if refresh fails. No loop risk. ✅
- **`getSupabase` import:** correctly imported from `./supabase` ✅
- **Token flow:** `refreshSession()` → `data.session.access_token` → new `Headers` from original `init?.headers` (preserves any other request headers) → retry with fresh Bearer ✅
- **Fallback path:** if `getSupabase()` returns null (auth disabled) or `data.session` is null after refresh → `auth.signOut()` as before ✅
- **`return res`** at the bottom still returns the original 401 response if auth is disabled (`!auth.authEnabled`) — correct, no change to that path ✅

### No regressions
- The three moved routes (`/status`, `GET /notifications`, `GET /events`) are functionally identical to their removed counterparts — pure copy-move, no logic changes.
- Existing auth-required endpoints unchanged.

Ship it. 🚀

PR #175, opened 2026-05-15
feat: require JWT auth on all API endpoints when Supabase is enabled (#162)

### LINKED ISSUE #162: Add authentication to the HTTP API to restrict access to IO only
## Problem

The local HTTP API currently has no authentication, meaning any process on the machine can send requests to it. Access should be restricted so only IO itself (or trusted local clients) can communicate with it.

## Proposed Solution

Add an authentication mechanism to the Express API so that unauthenticated requests are rejected. Options to consider:

- **Shared secret / API key** — IO generates or stores a secret token at startup; all API requests must include it in an `Authorization` header (e.g. `Bearer <token>`). The token could be stored in `~/.io/config.json` or as an environment variable.
- **Localhost-only binding + token** — Combine binding the server to `127.0.0.1` (already likely the case) with a token check for defense in depth.

## Acceptance Criteria

- [ ] Unauthenticated requests to the API return `401 Unauthorized`
- [ ] IO's internal clients (TUI, Telegram bot, squad workers) pass the token correctly
- [ ] The secret is not hardcoded — it is generated on first run or read from config
- [ ] Existing API functionality is unaffected for authenticated requests

michaeljolley:
Fixes #162

## Problem
The API had several endpoints placed before `requireAuth` that could be called without a valid JWT when Supabase auth was enabled.

### Server (src/api/server.ts)
- Move `requireAuth` middleware before ALL endpoints except `/health` and `/auth/config`
- Skills, inbox, status, notifications, and SSE events all now require valid JWT
- Write endpoints (POST/DELETE) were already behind auth — no change there

### Frontend (web/src/components/AppNav.vue)
- SSE EventSource now reconnects with token refresh on auth failure
- 3 retries with increasing backoff (2s, 4s, 6s) before giving up
- Properly cleans up stale connections before reconnecting

### Existing safeguard (from PR #165, unchanged)
- `apiFetch` retries once with `refreshSession()` before signing out on 401

## How it works together
1. Page loads → AppNav calls /status, /notifications → may get 401 (expired JWT)
2. `apiFetch` catches 401 → refreshes Supabase session → retries with new token → 200
3. SSE /events → may get 401 → `onerror` fires → refreshes token → reconnects
4. No session loss cascade

TSC clean, 72/72 tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ web-dist/
 docs/node_modules
 docs/.vitepress/dist
 docs/.vitepress/cache
+data/
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -98,7 +98,10 @@ export async function startApiServer(): Promise<void> {
     });
   });
 
-  // Skills read endpoints — public (no auth required; read-only local filesystem)
+  // Apply auth middleware — all routes below require a valid JWT
+  api.use(requireAuth);
+
+  // Skills read endpoints
   api.get("/skills", (_req: Request, res: Response) => {
     try {
       const skills = listSkills();
@@ -130,7 +133,7 @@ export async function startApiServer(): Promise<void> {
     }
   });
 
-  // Inbox read endpoints — public (nav badge needs count without auth timing issues)
+  // Inbox read endpoints
   api.get("/inbox/count", (_req: Request, res: Response) => {
     try {
       const count = countInboxEntries();
@@ -151,12 +154,12 @@ export async function startApiServer(): Promise<void> {
     }
   });
 
-  // Status endpoint — public (non-sensitive version/uptime info)
+  // Status endpoint
   api.get("/status", (_req: Request, res: Response) => {
     res.json({ version: IO_VERSION, uptime: process.uptime() });
   });
 
-  // Notifications read endpoint — public (local-only app, nav badge needs data without auth race)
+  // Notifications endpoint
   api.get("/notifications", (_req: Request, res: Response) => {
     try {
       const unreadOnly = _req.query.unread === "true";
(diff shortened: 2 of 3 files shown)


michaeljolley:

TSC clean, 72/72 tests pass. No `data/io.db` in diff.

### server.ts — requireAuth placement ✅
Single `api.use(requireAuth)` at line 102 — confirmed via grep. Placed after `/health` and `/auth/config` (the two pre-auth setup routes), before all other endpoints.

No duplicate `requireAuth` registrations. Old comment "Apply auth middleware to all subsequent routes" removed along with the previous placement. Clean. ✅

**Note:** This reverses the squad decision from PR #165 (public routes for `/status`, `/notifications`, `/events`). The new squad decision ([2026-05-15 11:17:36]) explicitly supersedes that — Michael confirmed all endpoints must be auth-gated when Supabase is enabled; the apiFetch retry-with-refresh (PR #165) handles the token-refresh race client-side without needing public routes. Load-bearing context acknowledged and correctly applied.

### AppNav.vue — SSE reconnection ✅
**Retry bound:** `retries >= 3` guard fires when retries=3 → max 3 reconnection attempts (at 2s, 4s, 6s intervals). Bounded. ✅

**Backoff:** `2000 * (retries + 1)` — linear backoff: 2s → 4s → 6s. Reasonable for an SSE connection behind auth. ✅

**Cleanup on error:**
- `es.close()` called immediately ✅
- `if (notificationSource === es) notificationSource = null` — stale reference guard: if `connectSSE()` was already called again before this error fires, the guard prevents clearing the newer connection ✅
- `onUnmounted` still calls `notificationSource?.close()` ✅

**Token refresh before reconnect:**
- `getSupabase()` + `refreshSession()` before setting the `setTimeout` ensures the new `authenticatedUrl()` call in the retry carries a fresh JWT ✅
- Wrapped in `try/catch { /* best effort */ }` — failure doesn't block reconnect attempt ✅

### .gitignore ✅
`data/` added. No `data/io.db` appears in the diff — the DB was not committed. ✅

Clean, well-structured PR. Ship it. 🚀

PR #178, opened 2026-05-15
feat(web): make web app mobile responsive (#171)

### LINKED ISSUE #171: Make web app mobile friendly
## Summary

The IO web app should be responsive and usable on mobile devices. Review layout, navigation, typography, and interactive elements to ensure a good experience on small screens.

michaeljolley:
Fixes #171

### AppNav — Mobile bottom tab bar
- Sidebar hidden below 768px (`hidden md:flex`)
- Fixed bottom nav with 8 icon+label tabs
- Active state: 2px cyan top-line indicator
- iOS safe area padding (`env(safe-area-inset-bottom)`)
- Notification/inbox badges preserved on mobile

### App.vue
- `pb-16 md:pb-0` on main content to clear bottom nav

### WikiView — Single-panel mobile
- Responsive flex direction (`flex-col md:flex-row`)
- List/content toggle with back button on mobile

### All views
- Outer padding: `p-3 sm:p-6` (reduced on mobile)
- Tables already have `overflow-x-auto` from prior work

Desktop layout fully unchanged. TSC clean, 72/72 tests pass.

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -16,7 +16,7 @@
 
     <AppNav />
 
-    <main class="flex-1 overflow-auto relative z-0">
+    <main class="flex-1 overflow-auto relative z-0 pb-16 md:pb-0">
       <!-- Subtle gradient highlight at the top edge of the content pane -->
       <div class="pointer-events-none absolute top-0 inset-x-0 h-[2px] bg-gradient-to-r from-transparent via-accent/[0.1] to-transparent z-10" aria-hidden="true"></div>
       <RouterView v-slot="{ Component }">
--- a/web/src/components/AppNav.vue
+++ b/web/src/components/AppNav.vue
@@ -116,40 +116,7 @@
               </button>
             </div>
           
-<!-- ── Mobile bottom navigation ── -->
-<nav
-  v-if="route.name !== 'login'"
-  class="md:hidden fixed bottom-0 inset-x-0 z-50 flex items-stretch border-t border-edge"
-  style="background: linear-gradient(180deg, rgba(8,13,22,0.93) 0%, rgba(5,8,15,0.98) 100%); backdrop-filter: blur(12px); padding-bottom: env(safe-area-inset-bottom)"
-  aria-label="Mobile navigation"
->
-  <RouterLink
-    v-for="item in navItems"
-    :key="item.to"
-    :to="item.to"
-    class="relative flex-1 flex flex-col items-center justify-center gap-0.5 py-2 px-1 min-w-0 transition-colors duration-150 touch-manipulation select-none"
-    :class="isActive(item.name) ? 'text-accent' : 'text-txt-muted'"
-    @click="item.name === 'notifications' ? onNotificationsClick() : item.name === 'inbox' ? onInboxClick() : undefined"
-    :aria-label="item.label"
-  >
-    <!-- Active top accent line -->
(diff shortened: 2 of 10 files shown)


michaeljolley:
**QA Review — WilyKat: REQUEST CHANGES** 🔴

TSC clean, 72/72 tests pass. Checklist items 1–7 all pass structurally — but the template nesting contains a blocking bug.

---

### 🔴 Blocking: Mobile nav gated behind `auth.authEnabled && auth.user`

The mobile `<nav>` was inserted inside `<template v-if="auth.authEnabled && auth.user">` in the footer block. This means the bottom nav only renders when Supabase auth is **both enabled and the user is signed in**.

Consequences:
1. **Non-Supabase deployments** (`authEnabled = false`): desktop sidebar is `hidden md:flex` (CSS-hidden on mobile), mobile nav never renders → **zero navigation on mobile**
2. **Signed-out users**: same result — navigated to login page, no way back if the router doesn't redirect correctly
3. **`collapsed = true` from localStorage** (secondary): mobile nav is also inside `<template v-if="!collapsed">`, so a user who had the sidebar collapsed on desktop won't see the mobile nav — also a latent bug, though less likely to be hit

The mobile nav is a structural app element, not footer content. It should be a top-level sibling of the desktop `<nav>`, not nested 3 levels deep in the footer's auth branch.

**Fix:** Move the `<nav class="md:hidden ...">` out of the footer templates entirely. The correct location is as a direct sibling of the desktop `<nav>` in the root template (similar to how `pb-16 md:pb-0` handles clearance in `App.vue`). Its own `v-if="route.name !== 'login'"` guard is all it needs.

---

### ✅ Everything else passes

| Check | Result |
|---|---|
| Sidebar `hidden md:flex` | ✅ `AppNav.vue` line 3 |
| Bottom nav `md:hidden` | ✅ `class="md:hidden fixed bottom-0 inset-x-0..."` |
| All 8 routes (`v-for="item in navItems"`) | ✅ Same array as desktop |
| `pb-16 md:pb-0` on `<main>` | ✅ `App.vue` |
| WikiView `flex-col md:flex-row` + back button | ✅ |
| Desktop layout preserved with `md:` prefixes | ✅ |
| `<meta name="viewport" content="width=device-width, initial-scale=1.0">` | ✅ `index.html` line 6 |
| `npx tsc --noEmit` | ✅ Clean |
| `npm test` | ✅ 72/72 pass |

Unblock this by relocating the mobile nav — everything else is solid. 🚀

michaeljolley:
Fixed: moved mobile nav outside the auth-gated block.

**Root cause**: The `</template>` closing the `<template v-if="auth.authEnabled && auth.user">` block was accidentally displaced by the mobile nav insertion, causing the nav to be nested inside the auth guard.

**Fix**: 
- Restored `</template>` after the sign-out button (properly closes the auth block)  
- Mobile `<nav>` is now a top-level sibling of the desktop `<nav>` via Vue 3 fragment support
- Gated only by `v-if="route.name !== 'login'"` — works in all deployments, auth states, and sidebar collapse states

TS: ✅ clean | Tests: 72/72 ✅

michaeljolley:
**QA Re-Review — WilyKat: APPROVED** ✅

Fix verified. Mobile `<nav class="md:hidden">` is now a top-level sibling of the desktop nav (line 155), outside both `v-if="!collapsed"` and `v-if="auth.authEnabled && auth.user"`. Only gated by `v-if="route.name !== 'login'"` — correct for all deployment modes.

Template tags balanced (4/4). TSC clean, 72/72 tests pass. Ship it. 🚀

PR #182, opened 2026-05-15
feat(web): auto-scroll to bottom when opening Chat tab (#180)

### LINKED ISSUE #180: Auto-scroll to bottom when opening Chat tab
## Summary
When navigating to the Chat tab in the web UI, the view should automatically scroll to the bottom so the user sees the most recent messages immediately.

## Expected Behavior
- Opening the Chat tab scrolls instantly to the bottom of the message list
- The user never lands mid-conversation with older messages visible first

## Current Behavior
The chat view loads at the top (or current scroll position), requiring the user to manually scroll down to see the latest messages.

michaeljolley:
Closes #180

Scrolls the chat message container to the bottom on mount and on route navigation so users always see the latest messages.

## Changes
- `onMounted`: scroll to bottom on initial load
- `watch(route.name)`: scroll to bottom whenever user navigates to the `/chat` route from another tab
- Existing scroll-on-new-message behavior (`watch(store.messages.length)`) unchanged

## Validation
- `npx tsc --noEmit` ✅
- `npm test` ✅ 72/72 
- `vite build` ✅

### DIFF
--- a/web/src/views/ChatView.vue
+++ b/web/src/views/ChatView.vue
@@ -135,13 +135,15 @@
 </template>
 
 <script setup lang="ts">
-import { ref, nextTick, watch, computed } from 'vue'
+import { ref, nextTick, watch, computed, onMounted } from 'vue'
+import { useRoute } from 'vue-router'
 import FluentIcon from '../components/FluentIcon.vue'
 import { useChatStore } from '../stores/chat'
 import { renderMarkdown } from '../lib/markdown'
 import { apiFetch, authenticatedUrl } from '../lib/api'
 
 const store = useChatStore()
+const route = useRoute()
 const input = ref('')
 const stopping = ref(false)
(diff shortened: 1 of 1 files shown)


michaeljolley:

TSC clean, 72/72 tests pass.

**Correctness:**
- `scrollToBottom()` wraps in `nextTick()` internally — the `onMounted` call is safe before messages are rendered; Vue defers the scroll until after the first DOM flush ✅
- Route watcher fires only when `name === 'chat'`, no spurious scrolls on other tabs ✅
- Watcher default (`immediate: false`) means it doesn't fire on setup — `onMounted` correctly handles initial load ✅
- Existing message-length watcher and manual scroll-up detection (`isScrolledUp`) untouched ✅

Minimal, targeted fix. Ship it. 🚀

PR #184, opened 2026-05-15
feat: unified feed merging inbox + notifications

michaeljolley:
## Summary
Consolidates the separate Inbox and Notifications systems into a single unified feed.

### Changes
- **Store**: New `unified_feed` table with `type` column (`deliverable` | `notification`), read/unread tracking, and delete support for all items. Data migration from both old tables.
- **API**: New `/feed` endpoints replacing separate `/inbox` and `/notifications` routes. Supports type filtering, read/unread, delete, mark-all-read.
- **Routing**: `notifyBackground()` writes `type=notification`, `shouldRouteToInbox()` writes `type=deliverable`. Fan-out preserved (notifications → Telegram+TUI+SSE).
- **TUI**: `/inbox` commands use unified feed store, show unread markers.
- **Web UI**: New `FeedView.vue` with filter tabs (All/Deliverables/Notifications), type icons (tray vs bell), read/unread tracking, delete for all items. Single nav item replaces two.
- **Tests**: 26 new tests for feed store (98 total, all passing).

### Migration
Automatic idempotent migration copies existing inbox_entries → deliverable, background_notifications → notification. Old tables preserved but no longer written to.

### Breaking Changes
- `/api/inbox` and `/api/notifications` endpoints removed (replaced by `/api/feed`)
- SSE event type changed from `notification` to `feed`
- `/inbox` and `/notifications` routes redirect to `/feed`


michaeljolley:

**Checks:**
- `cd web && npx vite build` — clean (115 modules, no warnings)
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**Stale imports:** `grep -rn 'store/inbox|store/notifications' src/` → no matches outside the legacy files themselves ✅

**feed.ts SQL:**
- `listFeedEntries`: dynamic WHERE with positional params array — correct. No string interpolation of user input. `id DESC` tiebreaker for deterministic same-second ordering ✅
- `countUnreadFeedEntries`: two separate prepared statements (with/without type) — clean ✅
- `markFeedEntryRead`: idempotent — returns true if row exists (already read) vs false if missing ✅
- All other functions: safe parameterized queries ✅

**db.ts migration:**
- `CREATE TABLE/INDEX IF NOT EXISTS` — idempotent ✅
- Data migration guarded by `io_state` flag — runs exactly once ✅
- try/catch — safe on fresh installs where old tables may not exist ✅
- `INSERT OR IGNORE` prevents duplicate rows on re-run ✅

**FeedView.vue FluentIcon bindings:**
- All 7 usages use `:paths=\'`...`'` (bound prop with template literal) — correct syntax per squad decision 2026-05-15 11:30:01 ✅

**Route ordering in server.ts:**
- `POST /feed/read-all` registered before `POST /feed/:id/read` — no Express route shadowing ✅

**Non-blocking observation:**
- `broadcastNotificationToSSE` now emits `type: 'feed'` (was `'notification'`). AppNav's SSE listener should handle the renamed event type. Worth a follow-up to confirm the bell counter still increments correctly on new feed entries.

PR #189, opened 2026-05-15
fix(web): resilient token retrieval prevents sporadic auth failures (#188)

### LINKED ISSUE #188: Sporadic auth failures on /status and /feed/count endpoints — sign out/in required to recover
## Description

The `/status` and `/feed/count` API endpoints sporadically begin returning failures. Refreshing the page does not resolve the issue. Signing out and back in always restores normal behavior.

## Steps to Reproduce

1. Use the IO web app normally
2. At some point (non-deterministic), `/status` and/or `/feed/count` start failing
3. Refreshing the page does not help
4. Signing out and signing back in resolves the issue

## Expected Behavior

Both endpoints should remain functional without requiring re-authentication.

## Actual Behavior

Endpoints fail sporadically. Only a full sign-out/sign-in cycle fixes it.

## Suspected Cause

Possible Supabase refresh token handling issue — the session token may be expiring and the client is not silently refreshing it before making API calls. When the token expires, requests fail until a fresh session is established via sign-in.

## Areas to Investigate

- Supabase session refresh logic in the API middleware/auth layer
- Whether the server-side Supabase client is calling `supabase.auth.getUser()` (validates against the server) vs. `supabase.auth.getSession()` (trusts local cache)
- Token expiry handling — are expired tokens being silently refreshed or just failing?
- Whether the issue is client-side (stale token sent) or server-side (token not validated/refreshed correctly)

michaeljolley:
Closes #188

## Root Cause
getAccessToken() read from a cached Pinia ref that could hold a stale expired token if Supabase's background auto-refresh silently failed.

## Fix
- getAccessToken() now calls supabase.auth.getSession() to always fetch the latest token from Supabase's internal state, updating cached refs as a side-effect
- apiFetch retry on 401 only signs out when refresh explicitly returns no session (not on network errors)
- authenticatedUrl() is now async to support the async token fetch
- All four call sites updated to await authenticatedUrl()

### DIFF
--- a/web/src/components/AppNav.vue
+++ b/web/src/components/AppNav.vue
@@ -252,11 +252,11 @@ onMounted(async () => {
   connectSSE()
 })
 
-function connectSSE(retries = 0) {
+async function connectSSE(retries = 0) {
   notificationSource?.close()
   notificationSource = null
 
-  const es = new EventSource(authenticatedUrl('/api/events'))
+  const es = new EventSource(await authenticatedUrl('/api/events'))
   notificationSource = es
 
   es.onmessage = (e) => {
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -7,7 +7,7 @@ import { getSupabase } from './supabase'
  */
 export async function apiFetch(input: string, init?: RequestInit): Promise<Response> {
   const auth = useAuthStore()
-  const token = auth.getAccessToken()
+  const token = await auth.getAccessToken()
 
   const headers = new Headers(init?.headers)
   if (token) {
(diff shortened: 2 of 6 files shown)


michaeljolley:

**Checks:**
- `cd web && npx vite build` — clean (115 modules)
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**All callers of `authenticatedUrl` properly awaited:**
- `AppNav.vue:259` — `await authenticatedUrl` ✅
- `AgentActivityView.vue:581` — `await authenticatedUrl` ✅
- `ChatView.vue:241` — `await authenticatedUrl` ✅
- `SquadsView.vue:458` — `await authenticatedUrl` ✅

**No synchronous `getAccessToken()` calls remaining:**
- `api.ts:10` and `api.ts:50` both `await auth.getAccessToken()` ✅
- Return type is now `Promise<string | null>` — TypeScript would catch any missed awaits ✅

**401 retry logic improvement:**
- Previous: any 401 → refresh attempt → if no session → signOut (would nuke session on transient network error)
- New: 401 → refresh → if session → retry; if no session AND no error → signOut; if network error → leave session intact ✅
- The `!error` guard is the right condition: only sign out when Supabase positively confirms no valid session ✅

**`getAccessToken()` implementation:**
- Calls `supabase.auth.getSession()` for a fresh token on every invocation ✅
- Falls back to cached `session.value?.access_token` when Supabase client unavailable ✅
- Early-return `null` when auth not enabled avoids unnecessary Supabase calls ✅
- Side-effect of updating `session.value` + `user.value` on success is benign (keeps store in sync) ✅

No FluentIcon changes in this PR — no syntax check needed.

PR #192, opened 2026-05-16
fix(web): avoid getSession() side-effect that wiped cached session (#191)

### LINKED ISSUE #191: bug: 401 on /status and /feed/count after login — getSession() side-effect clears cached session
## Description
After installing v0.21.0, `/api/status` and `/api/feed/count` fail with 401 errors once logged in. The response is `{"error":"Missing or invalid authorization"}`.

## Root Cause
PR #189 changed `getAccessToken()` to call `supabase.auth.getSession()` on every API call. In Supabase JS v2, `getSession()` can trigger internal state changes — if the session is expired or the internal lock times out, it can fire `onAuthStateChange` with `session = null`, which our handler uses to wipe `session.value`. The fallback then returns null, no Authorization header is sent, and the server returns 401.

## Steps to Reproduce
1. Install v0.21.0
2. Log in via Supabase auth
3. Observe that `/api/status` and `/api/feed/count` calls fail with `{"error":"Missing or invalid authorization"}`
4. Sign out and sign back in recovers

## Expected Behavior
Both endpoints return 200 OK with valid data.

## Fix
- Revert `getAccessToken()` to read from cached `session.value` (no `getSession()` call)
- Add proactive refresh when session is close to expiry (within 30s of `expires_at`)
- Keep the signOut guard from PR #189 (only sign out on explicit no-session, not network errors)

michaeljolley:
Closes #191

## Root Cause
`getAccessToken()` was calling `supabase.auth.getSession()` on every API call. In Supabase JS v2, `getSession()` can trigger an internal state change that fires `onAuthStateChange` with `session = null`, which clears `session.value`. The function then falls through to the cleared cache and returns null → no Authorization header → 401.

## Fix
- Read from cached `session.value` (authoritative source via `onAuthStateChange`)
- Proactively call `refreshSession()` only when within 30s of expiry (safe — no side-effects on auth state)
- Fall back to cached token if refresh fails (transient network error shouldn't sign out)
- Never call `getSession()` from `getAccessToken()`

### DIFF
--- a/web/src/stores/auth.ts
+++ b/web/src/stores/auth.ts
@@ -83,16 +83,23 @@ export const useAuthStore = defineStore('auth', () => {
   }
 
   async function getAccessToken(): Promise<string | null> {
-    if (!authEnabled.value) return null
-    const supabase = await getSupabase()
-    if (!supabase) return session.value?.access_token ?? null
-    const { data } = await supabase.auth.getSession()
-    if (data.session) {
-      session.value = data.session
-      user.value = data.session.user
-      return data.session.access_token
+    const cached = session.value
+    if (!cached?.access_token) return null
+
+    // Proactively refresh when within 30 seconds of expiry.
+    // Uses refreshSession() (safe) rather than getSession() which can fire
+    // onAuthStateChange with session=null and wipe our cached session ref.
+    if (cached.expires_at && cached.expires_at * 1000 - Date.now() < 30_000) {
+      try {
+        const supabase = await getSupabase()
+        if (supabase) {
+          const { data } = await supabase.auth.refreshSession()
+          if (data.session) return data.session.access_token
+        }
+      } catch { /* fall through to cached token */ }
     }
-    return session.value?.access_token ?? null
+
+    return cached.access_token
   }
 
   return { user, session, loading, initialized, authEnabled, init, signIn, signOut, getAccessToken }


michaeljolley:

The fix correctly removes the `getSession()` call that was causing side-effects in Supabase JS v2, replacing it with a read from the cached `session.value` (authoritative source via `onAuthStateChange`). The proactive refresh within 30s of expiry is a good safety net. The 401 retry in `apiFetch` remains as the backstop.

Build, vue-tsc, and all 98 tests pass.

michaeljolley:

**Checks:**
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**Fix analysis:**
The root cause (per squad decision 2026-05-16 02:23:38) is that Supabase JS v2's `getSession()` can fire `onAuthStateChange` with `session=null` during internal state reconciliation, wiping the cached `session.value` ref before the returned token can be used. PR #189 introduced exactly this pattern.

**New implementation is correct:**
- Reads `session.value` directly — populated exclusively by `onAuthStateChange`, the authoritative session source ✅
- `expires_at * 1000 - Date.now() < 30_000` proactive-refresh window is sensible ✅
- Uses `refreshSession()` (safe) rather than `getSession()` (triggers spurious `onAuthStateChange(null)`) ✅
- Falls through to cached token on refresh failure — graceful degradation ✅
- Early returns `null` when no cached token, avoiding unnecessary Supabase calls ✅

PR #194, opened 2026-05-16
fix(web): guard onAuthStateChange against spurious null sessions

### LINKED ISSUE #193: bug: No Authorization header sent — onAuthStateChange clears session on spurious null events
## Description
Production users on io.baldbeardedbuilder.com see 401 "Missing or invalid authorization" errors because the browser sends NO Authorization header on API requests (GET /api/status, /api/feed/count, etc).

## Root Cause
The `onAuthStateChange` handler in `web/src/stores/auth.ts` blindly sets `session.value = s` for ALL Supabase auth events. In Supabase JS v2, internal events (token refresh failure, lock timeout, state reconciliation) can fire with `session = null` on events other than `SIGNED_OUT`. This wipes `session.value`, so `getAccessToken()` returns null and `apiFetch()` sends no Authorization header.

## Steps to Reproduce
1. Log in on io.baldbeardedbuilder.com
2. Observe network requests in browser DevTools
3. GET /api/status and GET /api/feed/count are sent without an Authorization header
4. Server returns `{"error":"Missing or invalid authorization"}`

## Fix (two-pronged)
1. Guard `onAuthStateChange`: only clear session on explicit `SIGNED_OUT` event
2. Recovery fallback in `getAccessToken()`: if cached session is null but auth is enabled, attempt `refreshSession()` to recover from localStorage refresh token

michaeljolley:
Closes #193

## Root Cause
The `onAuthStateChange` handler blindly set `session.value = s` for all Supabase auth events. In Supabase JS v2, internal events (token refresh failure, lock timeout, state reconciliation) can fire with `session = null` on events other than `SIGNED_OUT`. This wiped the cached session, causing `getAccessToken()` to return null → no Authorization header → 401.

## Fix
**1. Guard onAuthStateChange** — only clear `session.value` on explicit `SIGNED_OUT` events. Other null-session events are silently ignored, preserving the cached session.

**2. Recovery fallback in getAccessToken()** — when `session.value` is null but `authEnabled` is true, attempt `refreshSession()` to recover using the refresh token still in localStorage.

## What changed
- `web/src/stores/auth.ts` — 1 file, +26/-14

## Verification
- `npm run build` ✅
- `vue-tsc --noEmit` ✅
- `npm test` ✅ (98 tests)

### DIFF
--- a/web/src/stores/auth.ts
+++ b/web/src/stores/auth.ts
@@ -34,10 +34,18 @@ export const useAuthStore = defineStore('auth', () => {
     // transiently return null while a background token-refresh is in flight.
     await new Promise<void>((resolve) => {
       supabase.auth.onAuthStateChange((event, s) => {
-        session.value = s
-        user.value = s?.user ?? null
-        // INITIAL_SESSION fires exactly once at startup (session may be null
-        // when the user is not logged in — that is the correct outcome too)
+        if (s) {
+          // Accept any valid session regardless of event type
+          session.value = s
+          user.value = s.user ?? null
+        } else if (event === 'SIGNED_OUT') {
+          // Only clear on explicit sign-out — other null-session events
+          // (TOKEN_REFRESHED failure, internal lock timeout, etc.) must not
+          // wipe a valid cached session.
+          session.value = null
+          user.value = null
+        }
+        // INITIAL_SESSION fires exactly once at startup
         if (event === 'INITIAL_SESSION') {
           resolve()
         }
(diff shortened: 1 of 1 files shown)


michaeljolley:

Two-pronged fix correctly addresses the root cause:
1. **onAuthStateChange guard** — only `SIGNED_OUT` can clear the cached session. Other null-session events from Supabase JS v2 internals (TOKEN_REFRESHED failure, lock timeout) are now ignored, preventing the spurious session wipe.
2. **getAccessToken recovery** — if the cache IS wiped despite the guard, `refreshSession()` recovers from the localStorage refresh token. Safe — no `getSession()` call (per decision log).

Build, vue-tsc, and all 98 tests pass.

michaeljolley:

**Checks:**
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**Fix 1 — `onAuthStateChange` guard:**
- Previous: any `onAuthStateChange` event with null session wiped `session.value`
- New: only `SIGNED_OUT` clears the cache; all other null-session events (TOKEN_REFRESHED failure, internal lock timeout) are ignored ✅
- Correctly handles the Supabase JS v2 behavior where internal state reconciliation can emit null-session events that shouldn't be treated as logout ✅

**Fix 2 — `getAccessToken()` recovery path:**
- When cache is null but `authEnabled`, attempts `refreshSession()` as a safety net ✅
- Uses `refreshSession()` (safe), NOT `getSession()` — fully compliant with squad decision 2026-05-16 02:23:38 ✅
- Falls through to `null` on failure — no hard error, no infinite retry ✅

**Defense in depth:** Both fixes together handle the root cause (Fix 1 prevents spurious clearing) AND add resilience for edge cases where the cache still ends up empty (Fix 2 recovers). Correct layered approach.

PR #196, opened 2026-05-16
fix(web): defer AppNav API calls until auth is initialized

### LINKED ISSUE #195: bug: /status and /count fire before auth ready, /count called twice
## Description
AppNav mounts unconditionally (even on login page) and its `onMounted` fires `/api/status` and `/api/feed/count` before the router guard finishes `auth.init()`. This causes 401 errors. The 401-retry in `apiFetch` then triggers a second `/feed/count` call that succeeds.

## Fix
Add `await auth.init()` and a user check at the top of AppNav `onMounted` so API calls only fire once auth is settled.

michaeljolley:
Closes #195

## Problem
AppNav mounts before the router guard finishes `auth.init()`, so `/api/status` and `/api/feed/count` fire with no token → 401. The 401-retry in `apiFetch` then triggers a second `/feed/count` call.

## Fix
Add `await auth.init()` + user-check at the top of `onMounted`:
- Defers all API calls and SSE until auth state is settled
- Skips authenticated calls entirely on the login page (no user)
- Eliminates both the timing 401s and the duplicate `/count` call

## Changed
- `web/src/components/AppNav.vue` — +8 lines

### DIFF
--- a/web/src/components/AppNav.vue
+++ b/web/src/components/AppNav.vue
@@ -356,6 +356,14 @@ onMounted(async () => {
     if (stored !== null) collapsed.value = stored === 'true'
   } catch { /* ignore */ }
 
+  // Wait for auth to be fully initialized before making API calls.
+  // AppNav mounts before the router guard finishes auth.init(), so
+  // without this guard, requests fire with no token and get 401s.
+  await auth.init()
+
+  // Skip authenticated API calls when on the login page
+  if (auth.authEnabled && !auth.user) return
+
   try {
     const res = await apiFetch('/api/status')
     if (res.ok) {


michaeljolley:

Minimal, targeted fix. `await auth.init()` deduplicates via the existing `initPromise` so there's no double-init cost. The user-check skip on login page is correct — no point hitting authenticated endpoints when not logged in. This eliminates both the timing 401s and the retry-driven duplicate /count call.

michaeljolley:

**Checks:**
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**Fix analysis:**
- `await auth.init()` defers all API calls until auth state is settled ✅
- `auth.init()` is idempotent: guarded by `initialized.value` + `initPromise` deduplication — calling from both AppNav and the router guard won't cause a double-init race ✅
- `if (auth.authEnabled && !auth.user) return` — correctly skips authenticated calls on the login page / when not signed in. Does not skip when auth is disabled (non-Supabase deployment), so status/count still load for unauthenticated setups ✅
- Eliminates the timing window where AppNav mounted and fired `/api/status` + `/api/feed/count` before the router guard completed auth init ✅

PR #198, opened 2026-05-16
fix: resolve releases.json path for both dev and production

### LINKED ISSUE #197: API /releases endpoint returns empty array despite GitHub releases existing
### Description
The  API endpoint always returns , causing the UI to display "No release notes found" even when releases and release notes exist in GitHub.

### Steps to Reproduce
1. Navigate to the releases/changelog section of the IO UI
2. Observe "No release notes found" message

### Expected Behavior
The endpoint should return the list of GitHub releases with their associated release notes.

### Impact
Users cannot view release notes through the UI.

michaeljolley:
Closes #197

## Root Cause
`loadReleases()` reads from `path.join(__dirname, '../releases.json')`. In dev (tsx), `__dirname` = `src/api/` → resolves to `src/releases.json` ✅. In production (npm install), `__dirname` = `dist/api/` → resolves to `dist/releases.json` which doesn't exist (tsc doesn't copy .json, `files` didn't include it) → catch silently returns `[]`.

## Fix
1. `loadReleases()` checks two candidate paths: `../releases.json` then `../../src/releases.json`
2. Added `src/releases.json` to package.json `files`
3. Refreshed releases data to include v0.21.0-v0.21.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   },
   "files": [
     "dist/**/*.js",
+    "src/releases.json",
     "web-dist/**/*",
     "README.md"
   ],
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -26,13 +26,20 @@ const WEB_DIST = path.resolve(__dirname, "../../web-dist");
 let releasesCache: unknown[] | null = null;
 function loadReleases(): unknown[] {
   if (releasesCache) return releasesCache;
-  try {
-    const raw = readFileSync(path.join(__dirname, "../releases.json"), "utf-8");
-    releasesCache = JSON.parse(raw) as unknown[];
-    return releasesCache;
-  } catch {
-    return [];
+  // Check both dev (src/api/ → src/) and production (dist/api/ → src/) paths
+  const candidates = [
+    path.join(__dirname, "../releases.json"),         // dist/api/ → dist/releases.json
+    path.join(__dirname, "../../src/releases.json"),  // dist/api/ → src/releases.json
+  ];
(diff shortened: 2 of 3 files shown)


michaeljolley:

Clean fix. The dual-path candidate approach mirrors the existing pattern in `paths.ts` for `resolveVersion()`. Adding `src/releases.json` to `files` ensures it ships with the npm package. The `existsSync` guard is correct since it's already imported.

michaeljolley:

**Checks:**
- `npx tsc --noEmit` — clean
- `npm test` — 98/98 pass

**Path resolution fix:**
- Two candidate paths tried in order: `dist/api/../releases.json` (compiled alongside output) then `dist/api/../../src/releases.json` (source tree) ✅
- `existsSync()` check before read — no exception on missing path, just falls through ✅
- `existsSync` correctly imported from `node:fs` (already in imports) ✅
- Cache still works correctly — first successful read is stored in `releasesCache` ✅

**package.json `files`:**
- `src/releases.json` added — ensures the file is included in `npm pack` / published package ✅
- Correct fix for production deploys that install from npm rather than from source

**releases.json data:**
- Updated with v0.21.0–v0.21.3 entries at the top ✅
- Older entries (v0.13.0 and earlier) pruned to stay within the 10-release fetch limit ✅

**Non-blocking observation:**
- The v0.21.1–v0.21.3 release bodies have duplicated `## What's Changed` sections (GitHub generated them twice). Cosmetic only — doesn't affect the feature.

PR #217, opened 2026-05-18
feat: use LLM to generate real character rosters for custom universes (#216)

### LINKED ISSUE #216: feat: LLM-generated character rosters for custom universes
## Problem

When `squad_create` is used with an unknown universe (e.g. "he-man", "futurama"), the system falls back to generic archetype names (Commander, Architect, Scout) instead of real characters from that universe.

## Solution

Use the Copilot SDK to dynamically generate a roster of real, canonical characters with accurate personalities for any pop-culture universe not in the built-in list.

## Acceptance Criteria
- `squad_create` with `universe: "he-man"` produces agents named He-Man, Teela, Man-At-Arms, Orko, etc.
- Any reasonable pop-culture universe produces real characters
- Built-in universes bypass the LLM call entirely (no latency penalty)
- Graceful fallback to archetypes if LLM fails

## Implementation
PR #216 (pending)

michaeljolley:
## Summary

When `squad_create` receives an unknown universe name (not in built-in or well-known lists), the system now uses the Copilot SDK to ask the LLM for 8 real canonical characters with accurate personalities, rather than falling back to generic archetypes.

### How it works

1. `generateUniverseRoster(input)` is called in the `squad_create` handler before `createSquad()`
2. If the universe resolves to a known roster → returns immediately (no LLM call)
3. If it resolved to archetypes → fires a one-shot LLM prompt asking for 8 real characters
4. Parses the JSON response, validates it, and registers it in the UNIVERSES array
5. If the LLM fails (timeout, parse error, network) → gracefully degrades to archetype fallback

### Examples

- `universe: "he-man"` → He-Man, Teela, Man-At-Arms, Orko, Skeletor, She-Ra, Battle Cat, Sorceress
- `universe: "futurama"` → Fry, Leela, Bender, Professor Farnsworth, Amy, Zoidberg, Hermes, Nibbler

### Technical details

- Dynamic `import("./client.js")` avoids circular deps (universes.ts doesn't statically import client)
- Session destroyed in `finally` block to prevent leaks
- LLM timeout: 30s (generous for a simple JSON generation task)
- All 131 tests pass, tsc clean

Closes #216

### DIFF
--- a/src/copilot/tools.ts
+++ b/src/copilot/tools.ts
@@ -4,7 +4,7 @@ import { execSync, execFileSync } from "child_process";
 import { readFileSync, writeFileSync, readdirSync, statSync, existsSync, mkdirSync } from "fs";
 import { join, dirname, resolve, sep } from "path";
 import { homedir } from "os";
-import { UNIVERSES, getOrCreateUniverse } from "./universes.js";
+import { UNIVERSES, getOrCreateUniverse, generateUniverseRoster } from "./universes.js";
 import { createFeedEntry } from "../store/feed.js";
 import { validateCron, nextRun } from "./cron.js";
 import {
@@ -427,6 +427,9 @@ export function createTools(deps: ToolDeps) {
     }),
     handler: async ({ slug, name, project_path, universe }) => {
       try {
+        if (universe) {
+          await generateUniverseRoster(universe);
+        }
         deps.createSquad(slug, name, project_path, universe);
         const squad = deps.getSquad(slug);
         const universeName = squad?.universe ? getOrCreateUniverse(squad.universe).name : "random";
--- a/src/copilot/universes.ts
+++ b/src/copilot/universes.ts
@@ -449,6 +449,71 @@ export function getOrCreateUniverse(input: string): Universe {
   return custom;
 }
 
+
+
+/**
+ * Generate a universe roster using the LLM for unknown universes.
+ * Falls back to archetype characters if the LLM call fails.
+ * For known/well-known universes, returns them directly without an LLM call.
+ */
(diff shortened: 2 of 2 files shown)


michaeljolley (commented):
LGTM — approved by Snarf (team lead). Clean implementation with graceful degradation. Dynamic import avoids circular deps, session cleanup in finally, archetype fallback if LLM fails. No impact on built-in universes. 131/131 tests pass.

michaeljolley (commented):
LGTM — approved by WilyKat (QA). Implementation is correct and all the tricky edge cases are handled.

**Verified:**
- LLM only fires for unknown universes: getOrCreateUniverse() runs first, isArchetype check gates the LLM call — known/well-known universes return early ✅
- Archetype detection characters[0]?.name === "Commander" is reliable — Commander is always the first generated archetype (line 404 of universes.ts) ✅
- JSON parsing strips markdown fencing from LLM responses ✅
- session.destroy() in finally block with best-effort try/catch ✅
- Graceful fallback: any error → logs + returns existing (archetype) ✅
- No circular dependency: client.ts does not import from universes.ts; dynamic import defers resolution to runtime ✅
- Archetype entry replaced in UNIVERSES via splice+push so nextCharacter finds the real roster ✅
- Characters capped at 8 with .slice(0, 8) to match universe pool size ✅
- String(c.name)/String(c.personality) coercion prevents LLM returning non-strings from crashing ✅

Non-blocking: the isArchetype heuristic would misfire if a real universe first character happened to be named "Commander" (e.g. a military franchise). More robust check would compare existing.id against slugify(input) to detect the archetype-generated path. Not worth blocking.

PR #214, opened 2026-05-18
feat: support free-form custom universes for squad creation (#213)

michaeljolley:
## Summary

Removes the hardcoded universe enum restriction from `squad_create`, allowing any free-form string as the universe theme.

### How it works

1. **Known universes** (12 total now): a-team, transformers, thundercats, gi-joe, aliens, ghostbusters + NEW: tmnt, star-wars, star-trek, lord-of-the-rings, the-office, parks-and-rec
2. **Name matching**: Input is matched against both IDs and display names (case-insensitive), so `"Teenage Mutant Ninja Turtles"` resolves to the TMNT roster
3. **Archetype fallback**: Unknown universes get auto-generated character archetypes flavored with the universe name (Leader, Strategist, Engineer, Scout, etc.)

### Changes
- `src/copilot/universes.ts`: 6 new universe rosters + `getOrCreateUniverse()` with tiered resolution
- `src/copilot/tools.ts`: Schema from `.enum(...)` → `.string().optional()`
- `src/store/squads.ts`: Uses `getOrCreateUniverse` instead of fallback-to-random

### Testing
- 131/131 tests pass
- tsc clean

Closes #213

PR #218, opened 2026-05-18
fix: re-register custom universe on addSquadAgent after daemon restart (#215)

### LINKED ISSUE #215: bug: custom universe characters lost after daemon restart
## Problem

When a custom universe (not one of the 6 original hardcoded ones) is used, the universe roster is registered in the runtime `UNIVERSES` array via `getOrCreateUniverse()`. After a daemon restart, this runtime state is lost. When `addSquadAgent` calls `nextCharacter(universeId, usedNames)`, it calls `getUniverse(universeId)` which returns `undefined` for the custom universe, causing an error.

## Fix

In `addSquadAgent` (src/store/squads.ts), call `getOrCreateUniverse(universeId)` before `nextCharacter(universeId, usedNames)` to ensure the universe is re-registered in the runtime array.

## Impact

Any squad using a custom universe (TMNT, Star Wars, Parks & Rec, etc.) will fail to add agents after daemon restart.

## Origin

Flagged by WilyKat during PR #214 review.

michaeljolley:
## Summary

One-line fix: calls `getOrCreateUniverse(universeId)` before `nextCharacter()` in `addSquadAgent` to ensure custom/well-known universes are re-registered in the runtime UNIVERSES array after a daemon restart.

**Root cause:** Custom universe rosters are stored in a runtime array that's lost on restart. The universe ID is persisted in SQLite, but `nextCharacter()` → `getUniverse()` couldn't find it in the empty runtime array.

**Fix:** `getOrCreateUniverse()` handles the re-registration (well-known lookup or archetype generation), making the universe available for `nextCharacter()`.

131/131 tests pass.

Closes #215

### DIFF
--- a/src/store/squads.ts
+++ b/src/store/squads.ts
@@ -121,6 +121,9 @@ export function addSquadAgent(
       .run(universeId, squadSlug);
   }
 
+  // Ensure universe is registered in runtime array (handles restart for custom universes)
+  getOrCreateUniverse(universeId);
+
   const existingAgents = listSquadAgents(squadSlug);
   const usedNames = existingAgents.map((a) => a.character_name);
   const character = nextCharacter(universeId, usedNames);


michaeljolley (commented):
LGTM — approved by Snarf (team lead). One-line fix correctly re-registers the universe before character lookup. 131/131 tests pass.

michaeljolley (commented):
LGTM — approved by WilyKat (QA). This is the exact one-liner fix I flagged during the PR #214 review.

getOrCreateUniverse(universeId) is already imported in squads.ts, so no new dependencies. Placed correctly: before nextCharacter(), after the universe is set on the squad. On restart: well-known universes get pushed back into UNIVERSES, custom universes regenerate archetypes (LLM path is async-only via squad_create, so archetypes are the correct sync fallback here). nextCharacter() now always finds the universe. Minimal, correct, no side effects. ✅

PR #220, opened 2026-05-18
fix: render multi-turn TUI responses as separate message bubbles (#219)

### LINKED ISSUE #219: TUI: Multi-turn responses append to same message bubble instead of rendering as separate messages
## Problem

When IO sends multiple messages in a single response turn, subsequent messages are appended to the same message bubble rather than appearing as new, separate messages. This makes the TUI output hard to read — sentences and paragraphs pile up in one growing block instead of flowing as distinct messages.

## Expected Behavior

Each discrete message IO sends should appear as its own message entry in the TUI, visually separated from prior messages in the same turn.

## Actual Behavior

New content is prepended/appended to the existing message string, making it look like one continuous, growing message rather than a natural conversation flow.

## Steps to Reproduce

1. Trigger any IO response that results in multiple message segments being sent (e.g., a response that includes an introductory line followed by a follow-up)
2. Observe the TUI — the messages concatenate into a single bubble instead of appearing as separate entries

## Notes

- Affects the TUI channel
- Likely a rendering/streaming issue in how the TUI handles incremental assistant message output

michaeljolley:
Fixes #219

Each message segment in a multi-turn response now renders as its own entry in the TUI rather than appending to a single growing bubble.

## What changed

In `src/tui/index.ts`, the `done=true` callback path previously called `rl.prompt()` inline, which meant:
- The prompt fired prematurely between messages
- `accumulated` was never reset, so subsequent messages appended to the previous content

**Fix:**
- When `done=true` fires, finalize and print the current bubble, then reset `accumulated` and `firstChunk` for the next message
- Move `rl.prompt()` to after `await messageHandler()` resolves — runs once, after all turns are complete

Single-message responses are unaffected: one `done=true` call prints the bubble, promise resolves, prompt fires. Multi-turn responses now print each message as its own line.

### DIFF
--- a/src/tui/index.ts
+++ b/src/tui/index.ts
@@ -279,15 +279,24 @@ export async function startTui(): Promise<void> {
         }
 
         if (done) {
+          // Finalize the current message as its own line, then reset state
+          // so the next message in a multi-turn response starts a fresh bubble.
           clearLine();
-          process.stdout.write(accumulated + "\n");
-          rl.prompt();
+          if (accumulated) {
+            process.stdout.write(accumulated + "\n");
+          }
+          accumulated = "";
+          firstChunk = true;
         } else {
           accumulated += text;
           clearLine();
           process.stdout.write(accumulated);
         }
       });
+      // Restore the prompt once the handler promise fully resolves (i.e. after
+      // all turns are done), rather than inside the done callback so that
+      // multi-turn responses don't render a premature prompt between messages.
+      rl.prompt();
     } catch (err) {
       clearLine();
       console.error(


michaeljolley (commented):
LGTM — approved by Snarf (team lead). Clean fix: resetting accumulated/firstChunk on each done=true ensures separate bubbles, and moving rl.prompt() outside the stream callback prevents premature prompt rendering. No concerns.

michaeljolley (commented):
LGTM — approved by WilyKat (QA). All three changes are correct: resetting accumulated+firstChunk on done=true ensures each turn gets a fresh bubble; moving rl.prompt() outside the stream callback prevents a premature prompt between multi-turn messages; the if (accumulated) guard cleanly skips blank-line writes for tool-call-only turns. No logic issues. ✅

PR #241, opened 2026-05-24
fix(web): ordered list rendering with correct close tags (#240)

### LINKED ISSUE #240: Bug: Ordered lists with repeated '1.' markers don't render with incrementing numbers
In the web markdown renderer (`web/src/lib/markdown.ts`), ordered lists written with repeated `1.` markers:

```markdown
1. something
1. something else
1. something altogether different
```

Should render as auto-incrementing numbers (1, 2, 3) but instead appear malformed because `closeList()` always emits `</ul>` — even when the list is an `<ol>`. This produces invalid HTML that browsers cannot render correctly.

**Root cause**: The `inList` boolean doesn't distinguish between ordered and unordered lists. When an ordered list opens with `<ol>`, `closeList()` incorrectly closes it with `</ul>`.

**Expected**: Standard Markdown behavior — repeated `1.` in an ordered list auto-increments displayed numbers via browser-native `<ol>` rendering.

michaeljolley:
Fixes #240 — ordered lists with repeated `1.` markers now render with auto-incrementing numbers.

## Root Cause
`closeList()` always emitted `</ul>`, even when the open list was an `<ol>`. The `inList` boolean didn't distinguish list types.

## Fix
- `inList: boolean` → `listType: 'ul' | 'ol' | null`
- `closeList()` now emits the correct closing tag based on what's open
- UL/OL handlers check for mismatched open lists and close them first

Minimal 6-line change. vue-tsc clean.

### DIFF
--- a/web/src/lib/markdown.ts
+++ b/web/src/lib/markdown.ts
@@ -161,14 +161,14 @@ export function renderMarkdown(md: string): string {
   const lines = md.split('\n')
   const out: string[] = []
   let inCode = false
-  let inList = false
+  let listType: 'ul' | 'ol' | null = null
   let inTable = false
   let codeLines: string[] = []
   let codeLang = ''
   let tableLines: string[] = []
 
   function closeList() {
-    if (inList) { out.push('</ul>'); inList = false }
+    if (listType) { out.push(`</${listType}>`); listType = null }
   }
 
   function closeTable() {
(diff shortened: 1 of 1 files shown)


michaeljolley (commented):
LGTM — approved by Snarf (squad lead). Clean minimal fix, correct close tag for both list types.

PR #252, opened 2026-05-24
feat(web): add blockquote rendering to markdown renderer (#251)

### LINKED ISSUE #251: Add blockquote rendering support to markdown renderer
The custom markdown renderer (web/src/lib/markdown.ts) has no support for `>` blockquote syntax — a standard Markdown feature that IO's own responses frequently use. Currently renders as raw text. Should add blockquote parsing that emits `<blockquote>` elements with appropriate styling.

michaeljolley:
Closes #251.

Adds `>` blockquote parsing to the custom markdown renderer. Consecutive `>` lines are grouped into a single `<blockquote>` element. Inline formatting (bold, italic, links) is preserved within blockquotes. The block closes cleanly on blank lines, headings, `---`, or EOF.

**Styling:** `border-l-2 border-txt-muted pl-3 my-2 text-txt-muted italic` — matches the dark UI theme.

**Tests:** 6 new test cases covering single-line, multi-line, inline formatting, close-on-blank-line, bare `>`, and EOF flush. All 47 tests pass.

### DIFF
--- a/web/src/lib/markdown.test.ts
+++ b/web/src/lib/markdown.test.ts
@@ -162,12 +162,51 @@ describe('renderMarkdown — fenced code blocks', () => {
 // ── blockquotes ───────────────────────────────────────────────────────────────
 
 describe('renderMarkdown — blockquotes', () => {
-  it('treats > prefix as plain text (blockquotes not implemented in this renderer)', () => {
-    // The renderer does not have a blockquote handler; > lines are rendered
-    // as paragraphs with the > entity-escaped. This test documents current behavior.
+  it('renders a single-line blockquote', () => {
     const html = renderMarkdown('> This is a quote')
-    contains(html, '&gt;')
+    contains(html, '<blockquote ')
+    contains(html, '</blockquote>')
     contains(html, 'This is a quote')
+    notContains(html, '&gt; This')
+  })
+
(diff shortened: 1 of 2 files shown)


michaeljolley (commented):
LGTM — approved by WilyKat.

Implementation correctly mirrors the existing `listType`/`closeList()` pattern. `closeBlockquote()` is wired at all the right exit points: headings, hr, blank lines, paragraph fallthrough, and EOF cleanup. Inline formatting inside blockquotes works correctly.

**One non-blocking behavior gap to note:** When a list follows a blockquote with no blank line in between (e.g. `> quote\n- item`), the list gets rendered *inside* the blockquote rather than after it, because the list-item handler doesn't call `closeBlockquote()`. Standard Markdown parsers would close the blockquote first. Practical impact is low (AI responses almost always have a blank line before lists), and the resulting HTML is structurally valid. Worth a follow-up fix + test case, but not blocking this PR.

Tests are solid — single-line, multi-line grouping, inline formatting, close-before-paragraph ordering assertion, bare `>` edge case, and EOF close all covered.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Clean state-machine extension matching the existing list pattern. Tests cover the key paths.

PR #257, opened 2026-05-24
feat: add send_to_inbox tool for direct inbox delivery (#254)

### LINKED ISSUE #254: Squads unable to write to IO inbox via `send_to_inbox` tool
## Problem

Squad agents (confirmed: `microsoft-powertoys-cmdpal`) are unable to deliver messages to Michael's IO inbox when instructed to use `send_to_inbox`. When the tool fails, agents fall back to embedding results in their task output instead of delivering them as separate inbox items.

The agent reported: _"The inbox API is unreachable (same issue as previous triage runs)"_ — suggesting this is a recurring problem, not a one-off.

## Impact

- Tasks that produce multiple discrete deliverables (e.g. one plan per issue) cannot be properly separated into individual inbox messages
- Michael must manually extract content from raw task output
- Other squads are likely affected

## Steps to Reproduce

1. Delegate a task to any squad instructing them to use `send_to_inbox` to deliver results
2. Observe that the agent reports the inbox API as unreachable and falls back to task output

## Expected Behavior

Squads should be able to write individual messages to Michael's IO inbox using `send_to_inbox` reliably.

## Tasks for Triage

- [ ] Investigate why `send_to_inbox` is unreachable from squad agent contexts
- [ ] Determine if this is an auth/permissions issue, a networking issue, or a missing tool registration
- [ ] Check if the inbox API endpoint is running and accessible from agent subprocesses
- [ ] Fix the root cause and verify squads can write to the inbox
- [ ] Add a smoke test or health check for inbox availability

/cc IO team

michaeljolley:
Fixes #254. Adds a send_to_inbox tool that squads can call directly to deliver content to Michael's IO inbox, rather than relying on keyword detection in task descriptions.

## Changes
- New `send_to_inbox` tool with `title`, `body`, and optional `squad_slug` parameters
- Creates a `deliverable` feed entry, optionally prefixed with `[squad_slug]`
- No dep changes needed — uses existing `createFeedEntry` import

## Tests
All 227 backend + 47 web tests pass.

### DIFF
--- a/src/copilot/tools.ts
+++ b/src/copilot/tools.ts
@@ -2072,7 +2072,23 @@ export function createTools(deps: ToolDeps) {
     },
   });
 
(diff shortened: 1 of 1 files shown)


michaeljolley (commented):
LGTM — approved by WilyKat.

Minimal, correct implementation. `send_to_inbox` follows the exact same pattern as the existing task-result delivery at line 554 (same `createFeedEntry` call, same `type: 'deliverable'`, same `[slug]` prefix convention). `skipPermission: true` is appropriate — this is a non-destructive, intentional write from the agent. `squad_slug` optional is the right call since not all inbox messages come from a squad context. CI green.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Direct inbox tool matches existing feed entry pattern, optional squad_slug keeps it flexible.

PR #258, opened 2026-05-24
fix(web): close blockquote before starting list (#253)

### LINKED ISSUE #253: Fix blockquote/list transition edge case in markdown renderer
When a list immediately follows a blockquote without a blank line separator, the list nests inside the blockquote instead of closing it. Fix: add closeBlockquote() calls in the UL and OL list-item handlers, matching the existing cross-list close pattern.

michaeljolley:
Fixes #253.

Adds `closeBlockquote()` calls in the UL and OL list-item handlers so a list immediately following a blockquote (no blank line) renders as a sibling element, not nested inside the blockquote.

This mirrors the existing cross-list-type transition pattern (`closeList()` when switching between ul/ol).

**Tests:** 2 new regression tests — blockquote→ul and blockquote→ol without blank line. All 49 tests pass.

### DIFF
--- a/web/src/lib/markdown.test.ts
+++ b/web/src/lib/markdown.test.ts
@@ -208,6 +208,29 @@ describe('renderMarkdown — blockquotes', () => {
     contains(html, '<blockquote ')
     contains(html, '</blockquote>')
   })
+
+  it('closes blockquote before unordered list with no blank line (regression #253)', () => {
+    const html = renderMarkdown('> quoted\n- item one\n- item two')
+    contains(html, '<blockquote ')
+    contains(html, '</blockquote>')
+    contains(html, '<ul ')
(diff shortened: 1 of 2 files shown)


michaeljolley (commented):
LGTM — approved by WilyKat.

Exactly the fix I flagged in the PR #252 review. Two surgical `closeBlockquote()` calls added at the UL and OL entry points, placed correctly *before* the `if (!listType)` open guard so the blockquote closes before any new list element is emitted. Regression tests for both UL and OL cases use position-assertion style (`bqClose < ulOpen`) matching the existing blockquote ordering test pattern. CI green.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Correct transition guard matching the existing cross-list pattern.

PR #259, opened 2026-05-24
fix(web): restore list indentation and numbering in wiki-content (#256)

### LINKED ISSUE #256: Feed: Markdown list rendering broken — no indentation, numbers reset to 1
## Bug

Markdown lists in the Feed tab are rendering incorrectly.

## Symptoms

- Ordered and unordered lists are no longer indented
- Ordered list item numbers all render as `1.` instead of incrementing (1, 2, 3...)

## Expected Behavior

Lists should render with proper indentation and sequential numbering, matching standard markdown rendering.

## Example

Given:
```md
1. First item
2. Second item
3. Third item
```

**Expected**: Indented, numbered 1 / 2 / 3  
**Actual**: No indent, all show as `1.`

michaeljolley:
Fixes #256.

Tailwind preflight resets `list-style: none` and `padding: 0` on `ol`/`ul`, stripping numbering and indentation from rendered markdown lists.

**Changes:**
- Added `.wiki-content ol/ul/li` overrides in `web/src/style.css` (global, covers Feed, Inbox, Wiki, AgentActivity views)
- Added equivalent `.floating-chat-md :deep(ol/ul/li)` rules in `FloatingChat.vue` scoped styles

The markdown renderer already emits correct `<ol><li>` / `<ul><li>` structure (fixed in #241/#252) — this fix ensures the browser renders the native list markers and indentation.

### DIFF
--- a/web/src/components/FloatingChat.vue
+++ b/web/src/components/FloatingChat.vue
@@ -305,4 +305,7 @@ async function sendMessage() {
 .floating-chat-md :deep(code) { font-family: 'JetBrains Mono', monospace; }
 .floating-chat-md :deep(a) { color: #22d3ee; }
 .floating-chat-md :deep(a:hover) { text-decoration: underline; }
+.floating-chat-md :deep(ol) { list-style-type: decimal; padding-left: 1.5em; margin: 0.5em 0; }
+.floating-chat-md :deep(ul) { list-style-type: disc; padding-left: 1.5em; margin: 0.5em 0; }
+.floating-chat-md :deep(li) { margin: 0.25em 0; }
 </style>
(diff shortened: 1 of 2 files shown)


michaeljolley (commented):
LGTM — approved by WilyKat.

Root cause is correct: Tailwind's regex scanner reliably picks up class names in static strings but can miss classes built inside template literals in .ts files — those utility classes get purged in production builds, leaving preflight's `list-style: none; padding: 0` in effect. Plain CSS overrides with `.wiki-content ul/ol/li` and `.floating-chat-md :deep(ol/ul/li)` are the right fix — not a Tailwind safelist, which would be brittle.

One side effect worth noting: `.wiki-content ul { padding-left: 1.5em }` has higher specificity (0,1,1) than `renderMarkdown`'s `pl-2` utility (0,1,0), so the CSS rule wins on padding. The result is actually better typography — `list-outside` + 1.5em indent is preferable to `list-inside` + 0.5rem. Not a problem, just a side effect to be aware of if someone diffs the rendered output. CI green.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Global CSS overrides for wiki-content lists are the right approach since multiple views share the class.

PR #260, opened 2026-05-24
feat: unify communication channels — inbox consolidation + send_notification (#255)

### LINKED ISSUE #255: Unify squad communication channels: inbox, notifications, and deliverables
## Problem

There is currently confusion around the multiple communication channels available in the IO app — the Feed tab shows both "Notifications" and "Deliverables", and there's a separate "Inbox". It's unclear how these differ, when squads should use each, and there's no consistent tool or pattern for squads and the orchestrator to reliably send content to either channel.

This leads to:
- Squads falling back to embedding deliverables in raw task output (see issue #254)
- Inconsistent metadata on inbox items (missing squad name, instance, context)
- No clear contract on what belongs in a notification vs. the inbox
- "Deliverables" as a separate concept adding confusion without clear value

---

### 1. Consolidate Deliverables into Inbox
- Remove "Deliverables" as a separate UI concept — all substantive content should go to the **Inbox**
- Every inbox item must include:
  - **Squad** name (e.g. `microsoft-powertoys-cmdpal`)
  - **Instance** (if applicable, e.g. `...-47939`)
  - **Task ID** (so it can be traced back)
  - **Subject / title**
  - **Full content / body**
  - Timestamp

### 2. Define Clear Channel Semantics
| Channel | Purpose | Example |
|---|---|---|
| **Inbox** | Substantive content, plans, reports, deliverables | "Here is the implementation plan for #47821" |
| **Notifications** | Short status updates, alerts, FYIs | "Task 558a0fb3 complete", "CI failed on PR #42" |

Notifications should be a single sentence — no walls of text. If it needs more than one sentence, it belongs in the inbox.

### 3. Consistent Tools / Patterns for All Squads & Orchestrator
- Provide a **first-class tool** (or pair of tools) that squads and the orchestrator can reliably call:
  - `send_to_inbox(subject, content, squad, instance?, task_id?)` — delivers a structured item to Michael's inbox
  - `send_notification(message)` — sends a short status notification to the feed
- These should be registered as tools available to all squad agents and the IO orchestrator
- Consider whether these should also be exposed as **skills** so any agent can install and use them
- The existing inbox API failure (issue #254) must be resolved as part of this work — the tools are useless if the underlying API is unreliable

### 4. Orchestrator Awareness
- IO itself should use these same tools/patterns when proactively delivering information to Michael
- Document the pattern clearly so new squads pick it up automatically (e.g. in squad onboarding context or a shared skill)

---

## Acceptance Criteria

- [ ] "Deliverables" section removed or merged into Inbox in the Feed UI
- [ ] Inbox items always show squad, instance (if applicable), task ID, subject, and content
- [ ] `send_to_inbox` tool works reliably from all squad agent contexts (fixes #254)
- [ ] `send_notification` tool works reliably from all squad agent contexts
- [ ] Both tools documented and available to all squads and the orchestrator
- [ ] Notifications are capped to short status messages in the UI (no long content)
- [ ] Existing squads updated or notified of the new pattern

michaeljolley:
Closes #255.

**Frontend changes (this commit):**
- Renamed 'Deliverables' tab → 'Inbox' in FeedView
- Updated `FeedEntry.type` from `'deliverable'` to `'inbox'`
- Added `squad_slug`, `instance_id`, `task_id` nullable fields to `FeedEntry` interface
- Inbox entries now display metadata badges when present: squad slug (accent), instance_id and task_id (muted, truncated to 8 chars) — shown in both flat and grouped views
- Empty state text updated to 'No inbox items'

**Backend changes (previous commits on this branch):**
- `send_notification` tool added
- Feed entry type renamed from 'deliverable' → 'inbox' in schema/API

Notifications remain as a separate feed for short status messages.

### DIFF
--- a/src/api/server.ts
+++ b/src/api/server.ts
@@ -135,7 +135,7 @@ export async function startApiServer(): Promise<void> {
   api.get("/feed/count", (req: Request, res: Response) => {
     try {
       const rawType = req.query.type;
-      const type = rawType === "deliverable" || rawType === "notification"
+      const type = rawType === "inbox" || rawType === "notification"
         ? (rawType as FeedEntryType)
         : undefined;
       const count = countUnreadFeedEntries(type);
@@ -149,7 +149,7 @@ export async function startApiServer(): Promise<void> {
   api.get("/feed", (req: Request, res: Response) => {
     try {
       const rawType = req.query.type;
-      const type = rawType === "deliverable" || rawType === "notification"
+      const type = rawType === "inbox" || rawType === "notification"
         ? (rawType as FeedEntryType)
         : undefined;
       const unreadOnly = req.query.unread === "true";
@@ -290,7 +290,7 @@ export async function startApiServer(): Promise<void> {
   api.post("/feed/read-all", (req: Request, res: Response) => {
     try {
       const rawType = req.query.type;
-      const type = rawType === "deliverable" || rawType === "notification"
+      const type = rawType === "inbox" || rawType === "notification"
         ? (rawType as FeedEntryType)
         : undefined;
       const marked = markAllFeedEntriesRead(type);
@@ -359,8 +359,8 @@ export async function startApiServer(): Promise<void> {
       source_type?: unknown;
       source_ref?: unknown;
     };
-    if (type !== "deliverable" && type !== "notification") {
-      res.status(400).json({ error: "type must be 'deliverable' or 'notification'" });
+    if (type !== "inbox" && type !== "notification") {
+      res.status(400).json({ error: "type must be 'inbox' or 'notification'" });
       return;
     }
     if (!title || typeof title !== "string" || title.trim() === "") {
--- a/src/copilot/scheduler.ts
+++ b/src/copilot/scheduler.ts
@@ -91,7 +91,7 @@ async function fireSchedule(schedule: SquadSchedule): Promise<void> {
   try {
     await delegateToAgent(squad.slug, prompt, (_taskId, result) => {
       if (shouldRouteToInbox(prompt)) {
-        createFeedEntry({ type: "deliverable", title: `[${squad.slug}] ${schedule.name}`, body: result });
+        createFeedEntry({ type: "inbox", title: `[${squad.slug}] ${schedule.name}`, body: result });
         console.error(`[io] Schedule ${schedule.id} result routed to inbox`);
         completeScheduleRun(run.id, 0);
       } else {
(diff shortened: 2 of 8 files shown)


michaeljolley (commented):
Changes requested by WilyKat: critical migration bug — the 'deliverable' → 'inbox' rename will silently break all existing databases.

## Root cause

`db.ts` uses a try/catch migration loop that swallows all errors. On any existing database the `unified_feed` table still has `CHECK(type IN ('deliverable', 'notification'))`. The migration statement:

```sql
UPDATE unified_feed SET type = 'inbox' WHERE type = 'deliverable'
```

throws `CHECK constraint failed` and is silently ignored. After that, every `createFeedEntry({ type: 'inbox', ... })` INSERT also throws the same error. Confirmed locally:

```
UPDATE FAILED: CHECK constraint failed: type IN ('deliverable', 'notification')
INSERT inbox FAILED: CHECK constraint failed: type IN ('deliverable', 'notification')
```

New installations work fine (the `CREATE TABLE IF NOT EXISTS` DDL has the correct constraint), but any upgrade silently kills inbox writes entirely.

## Fix required

SQLite doesn't support `ALTER TABLE ... MODIFY COLUMN`, so changing a CHECK constraint requires a table rebuild. The migration should be one atomic block:

```sql
ALTER TABLE unified_feed RENAME TO unified_feed_old;
CREATE TABLE unified_feed (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  type TEXT NOT NULL CHECK(type IN ('inbox', 'notification')),
  title TEXT NOT NULL,
  body TEXT NOT NULL,
  source_type TEXT,
  source_ref TEXT,
  squad_slug TEXT,
  instance_id TEXT,
  task_id TEXT,
  created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
  read_at DATETIME
);
INSERT INTO unified_feed
  SELECT id,
    CASE type WHEN 'deliverable' THEN 'inbox' ELSE type END,
    title, body, source_type, source_ref,
    NULL, NULL, NULL, created_at, read_at
  FROM unified_feed_old;
DROP TABLE unified_feed_old;
```

`better-sqlite3`'s `db.exec()` supports multi-statement SQL, so this can be a single migration string. The existing try/catch will handle 'already applied' on re-run (the RENAME will fail if `unified_feed_old` doesn't exist on a second run).

Everything else in this PR looks correct — type references, interface additions, frontend badges, `send_notification` tool. Just this migration path needs fixing before merge.

michaeljolley (commented):
LGTM — approved by WilyKat. Migration fix verified.

Table-rebuild sequence (RENAME → CREATE → INSERT with CASE type rename → DROP) is correct and confirmed working against an existing DB with the old CHECK constraint. 'deliverable' rows become 'inbox', new inserts work, and re-run is idempotent (RENAME fails silently on second pass, leaving the already-migrated table untouched). Two bonus indices on `(type, created_at)` and `(read_at, created_at)` are a clean addition. CI green.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Complete channel unification: table rebuild migration handles CHECK constraint correctly, send_notification tool follows established pattern, UI cleanly consolidates deliverables into inbox.

PR #265, opened 2026-05-25
feat: auto-complete squad instances when delegated task finishes (#261)

### LINKED ISSUE #261: Auto-complete squad instances when their delegated task transitions to done
## Problem

When a squad instance's delegated task completes (`status: done`), the instance remains open and idle. After 33 minutes of inactivity it gets auto-aborted, leaving behind a preserved worktree that requires manual cleanup.

This creates noise and busywork — the instance is functionally complete (work is done, PR is open) but the system treats it as abandoned rather than finished.

## Desired Behavior

When the task delegated to a squad instance transitions to `done`, the instance should automatically be marked complete (`squad_instance_complete`) — merging any decisions back to the master squad and cleaning up the worktree — without requiring manual intervention.

Alternatively, if auto-complete is not desirable (e.g. to allow follow-up tasks on the same instance), the inactivity timer should at minimum **not fire** when the instance's last task is in `done` state. An idle instance with no pending work is not the same as an abandoned one.

## Acceptance Criteria

- [ ] When a squad instance's delegated task reaches `done`, the instance is automatically completed (decisions merged, worktree cleaned up)
- [ ] OR: the auto-abort inactivity timer is suppressed when the instance's last task status is `done`
- [ ] No manual `squad_instance_complete` call required for the happy path (task completes → instance closes cleanly)
- [ ] Worktrees are not left behind for completed work

michaeljolley:
Closes #261.

**Problem**: Squad instances remained open/idle after their delegated task completed, eventually getting auto-aborted by the watchdog 30 minutes later.

**Fix**: 
- Added `autoCompleteInstance()` in `agents.ts` that fires immediately after `completeTask()` when the task has an `instance_id`
- Merges instance decisions back to master squad, cleans up worktree, sends notification
- Updated the instance watchdog to skip instances whose latest task status is `done` (avoids abort races)
- Guards: no-ops if instance is already in a terminal state (`done`/`failed`)

**Tests**: 5 new tests for auto-complete behavior + 1 new watchdog test. All 282 pass.

### DIFF
--- a/src/copilot/agents.ts
+++ b/src/copilot/agents.ts
@@ -43,6 +43,13 @@ import {
   getTask,
   cancelTask,
 } from "../store/tasks.js";
+import {
+  getInstance,
+  updateInstanceStatus,
+  mergeInstanceDecisions,
+} from "../store/instances.js";
+import { removeWorktree } from "../store/worktrees.js";
+import { createFeedEntry } from "../store/feed.js";
 import { SESSIONS_DIR } from "../paths.js";
 import { getUniverse } from "./universes.js";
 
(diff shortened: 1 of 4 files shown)


michaeljolley (commented):
LGTM — approved by WilyKat.

Auto-complete logic is correctly guarded: early returns for `done` and `failed` statuses prevent double-completion. Placement in `delegateToAgent` after `completeTask()` is correct per the squad decision — co-located with the task lifecycle, instanceId is available from closure, no extra polling needed.

Watchdog skip is correctly scoped to `latestTaskStatus === 'done'` via the same `COALESCE(completed_at, started_at) DESC` ordering used by the main staleness query. Only checking the latest task (not all tasks) is the right call — we only care about the most recent state.

One edge case noted but not blocking: if `autoCompleteInstance` crashes between `updateInstanceStatus('merging')` and `updateInstanceStatus('done')`, the instance gets stuck in `'merging'` permanently (watchdog queries `status = 'active'` only). However, `reconcileInstances()` on startup recovers this by failing any `'merging'` instance whose worktree is gone. Acceptable — the window is tiny and recovery exists.

Worktree path extraction regex (`/\.io-worktrees\/.*$/`) works correctly. CI green.

michaeljolley (commented):
LGTM — approved by Snarf (team lead). Clean auto-complete lifecycle integration. Guard on active-only instances prevents double-completion.

PR #271, opened 2026-05-25
fix: watchdog monitors stuck 'merging' instances (#267)

### LINKED ISSUE #267: bug: Watchdog should monitor 'merging' instances (crash recovery)
findStaleInstances() only queries status='active'. An instance that crashes mid-completion (stuck in 'merging') is invisible to the watchdog. reconcileInstances() covers startup, but a running daemon won't catch it. Add 'merging' to watchdog query with a shorter timeout (5min).

michaeljolley:
Closes #267. Extends instance watchdog to detect instances stuck in 'merging' state with a shorter 5-minute threshold, preventing zombie merging instances from lingering indefinitely.

**Changes:**
- `findStaleInstances()` now queries `status IN ('active', 'merging')`
- Merging instances use a separate threshold (default 5min vs 30min for active)
- Staleness clock for merging: last task `completed_at` (not task activity)
- Done-task skip logic only applies to active instances, not merging
- New `mergingThresholdMs` option in `InstanceWatchdogOptions`

**Tests:** 4 new tests for merging behavior + existing tests refactored. All 286 pass.

### DIFF
--- a/src/instance-watchdog.test.ts
+++ b/src/instance-watchdog.test.ts
@@ -30,7 +30,7 @@ beforeEach(() => {
 });
 
 describe("instance watchdog", () => {
-  describe("findStaleInstances", () => {
+  describe("findStaleInstances — active instances", () => {
     it("detects stale active instances with no task activity", () => {
       const db = getDb();
       db.prepare(`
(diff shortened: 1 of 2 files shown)


michaeljolley (commented):
QA approved by WilyKat — merging instances correctly monitored with shorter threshold, tests cover the new path. Key correctness checks: (1) merging branch hits `continue` before the done-task skip logic, so stuck-merging instances with a done latest task are correctly flagged (test 4 covers this); (2) created_at fallback for no-tasks merging is correct and tested (test 3); (3) active instance behavior unchanged. CI green.

PR #277, opened 2026-05-25
feat: MCP client auto-reconnect on server crash (#274)

### LINKED ISSUE #274: feat: MCP client auto-reconnect on server crash
When an MCP server subprocess crashes, the connection map retains a dead client. Errors surface safely as tool-level strings, but no automatic reconnect happens. Add reconnection logic on tool call failure (detect dead connection, clear from map, retry connect once).

michaeljolley:
Closes #274. Adds retry-once logic to callTool and listTools — on error, clears the dead connection and attempts one reconnect before surfacing the error.

**Strategy:** Single retry with connection eviction. If the first call fails, the cached client is removed from the map and a fresh connection is attempted. If the retry also fails, the error is thrown to the caller. This handles the common case of a crashed subprocess without masking persistent failures with infinite retries.

**Tests:** 4 new tests covering retry success, retry failure, listTools retry, and connection cleanup.

### DIFF
--- a/src/mcp/client.test.ts
+++ b/src/mcp/client.test.ts
@@ -1,0 +1,121 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { McpConnectionManager } from "./client.js";
+import type { McpServerConfig } from "./config.js";
+
+describe("McpConnectionManager — auto-reconnect", () => {
+  const testConfig: McpServerConfig = { name: "test-server", command: "echo", args: ["hello"] };
+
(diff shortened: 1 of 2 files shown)


michaeljolley (commented):
QA approved by WilyKat — retry-once pattern is correct, no infinite loops, error paths tested. connections.delete() before retry correctly forces getClient() to create a fresh transport. callTool wraps the second failure with a descriptive 'failed after reconnect' message; listTools lets the retry error propagate naturally (both safe). Test 4 guards against zombie connections remaining in the map after failure. CI green.

PR #281, opened 2026-05-25
feat: include squad wiki pages in instance snapshot and agent system prompt (#269)

### LINKED ISSUE #269: feat: Pass squad wiki/workflow context to squad instances at creation time
## Problem

When a squad instance is created via `squad_instance_create`, it receives a `context_snapshot` built from the master squad's decision log. However, **squad-specific workflow rules stored in the IO wiki** (branch naming conventions, PR labeling rules, build constraints, coding standards) are NOT included in that snapshot.

This causes instances to miss critical workflow rules. Real example: PT squad instances created branches that didn't follow the required `dev/mjolley/{name}` pattern because that rule was only in the wiki, not the decision log.

## Root Cause

`buildContextSnapshot()` in `src/store/instances.ts` pulls from:
- Squad decisions (`getDecisions()`)
- Active agent roster

It does **not** pull from:
- Wiki pages associated with the squad
- Any per-squad workflow/rules configuration

The wiki is a separate store (`src/wiki/`) with no integration into the instance lifecycle.

## Impact

- Instances operate without knowledge of squad-specific conventions
- Workflow violations (wrong branch names, missing labels, incorrect build steps) only caught after-the-fact
- Manual overhead: users must redundantly record rules in both the wiki AND the decision log

### Option 1: Auto-include wiki pages by squad slug
When creating an instance, scan `pages/squads/<slug>/` (or similar convention) in the wiki and include matching pages in `context_snapshot`.

**Pro**: Zero configuration, automatic  
**Con**: Requires a naming convention for wiki pages; may include irrelevant pages

### Option 2: Add a `workflow_rules` field to the squad record
A dedicated text field on the squads table (or a linked wiki page path) that gets explicitly concatenated into every instance snapshot.

**Pro**: Explicit, predictable  
**Con**: Yet another place to maintain rules (alongside wiki and decisions)

### Option 3: Squad-to-wiki-page linking
Allow squads to declare which wiki pages are "required reading" for instances (e.g., a `squad_wiki_pages` table or a JSON array on the squad row). At instance creation, those pages are fetched and included.

**Pro**: Flexible, opt-in per squad  
**Con**: Requires a new linking mechanism and maintenance

### Option 4: Surface a warning when wiki rules exist but aren't in decisions
Rather than auto-including wiki content, warn when a squad has wiki pages that might contain workflow rules not captured in the decision log.

**Pro**: Least invasive, encourages proper decision logging  
**Con**: Doesn't solve the problem directly; still requires manual action

## Open Questions for Michael

1. Which approach (or combination) makes the most sense?
2. Should wiki context be immutable in the snapshot (captured at creation) or live (re-read on each tool call)?
3. Is there an existing wiki page naming convention we should standardize on for squad-specific pages?
4. Should this also apply to the orchestrator's system prompt for squad agents (not just instances)?

## Acceptance Criteria (proposed)

- [ ] Squad instances receive relevant workflow rules at creation time
- [ ] Branch naming, PR conventions, and build rules are available to instance agents
- [ ] No duplicate maintenance burden (rules live in one place, flow everywhere needed)
- [ ] Existing squads with wiki-documented rules work without migration

michaeljolley:
Closes #269. Squad wiki pages at pages/squads/{slug}/ are now:
- Included immutably in instance context_snapshot at creation time
- Included in the agent system prompt for squad agents

This ensures squad-specific workflow rules (branch conventions, PR rules, etc.) are available to both instances and regular agents.

### DIFF
--- a/src/copilot/agents.ts
+++ b/src/copilot/agents.ts
@@ -52,6 +52,7 @@ import { removeWorktree } from "../store/worktrees.js";
 import { createFeedEntry } from "../store/feed.js";
 import { SESSIONS_DIR } from "../paths.js";
 import { getUniverse } from "./universes.js";
+import { readSquadWikiPages } from "../wiki/fs.js";
 
 export interface AgentInfo {
   slug: string;
@@ -531,6 +532,10 @@ async function getOrCreateAgentSession(
   const squad = getSquad(squadSlug)!;
   const client = await getClient();
   const decisions = getDecisionsSummary(squadSlug);
+  const wikiPages = readSquadWikiPages(squadSlug);
+  const wikiSection = wikiPages.length > 0
+    ? `\n\n## Squad Wiki\n${wikiPages.map(p => `### ${p.path}\n${p.content}`).join("\n\n")}`
+    : "";
 
   console.error(`[io] Agent ${agent.character_name}: using model "${model}" (agent tier: ${agentTier}, task tier: ${taskTier}, effective: ${effectiveTier})`);
 
@@ -601,7 +606,7 @@ ${agent.charter ?? "General-purpose agent. Handle tasks as they come."}
 - **Path**: ${squad.project_path}
 
 ## Past Decisions
-${decisions}${leadSection}
+${decisions}${leadSection}${wikiSection}
 
 ## Repository Hygiene
 Before you make ANY code changes, you MUST sync your working copy with the remote default branch and work from a fresh feature branch. This prevents the merge conflicts the team hit on PRs like #45.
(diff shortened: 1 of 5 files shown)


michaeljolley (commented):
QA approved by WilyKat — wiki pages correctly included in snapshot and system prompt, path filtering is safe, backward-compatible.

Path safety confirmed: slug is used only for string prefix matching against listPages() output (filesystem-derived paths), never for constructing file paths directly — traversal-crafted slugs produce a non-matching prefix, not file access.

Backward compat: snapshot format changed from bare array to {decisions, wiki?}. Only consumer is the agent system prompt builder in agents.ts, which is updated. No other callers parse this JSON.

Both getOrCreateAgentSession and getOrCreateSession updated correctly. Wiki section appended after leadSection (correct priority: decisions before wiki). Tests cover empty squad, single page, empty-content filter, multi-page, and snapshot inclusion. CI green.

PR #282, opened 2026-05-25
feat: Telegram image/file attachment support (#268)

### LINKED ISSUE #268: feat: File and image attachment support in prompts
## Summary

Support attaching images and files when prompting IO (via TUI and Telegram), and pass those attachments through to squad agents when delegating tasks. Vision-capable models should receive images directly.

---

### Open Questions
- **Input method**: How does the user attach a file in the terminal? Options:
  - `/attach <path>` command before or alongside a message
  - Drag-and-drop (paste file path from terminal emulator)
  - Inline syntax like `` in the prompt
- **Preview**: How is attached content shown before sending?
  - Images: thumbnail via sixel/kitty protocol (if terminal supports), otherwise just filename + size
  - Files: first N lines preview, or just metadata (name, size, mime type)
- **Size limits**: Max file size? Max number of attachments per message?

---

### Expected Behavior
- Telegram natively sends images/files as message attachments — IO should detect and handle these
- Photos: download via Telegram Bot API `getFile()`, store temporarily
- Documents: download, detect mime type, decide how to handle (text files inline, images as vision, binaries as references)
- Voice messages / audio: out of scope for v1? Or transcribe via Whisper?

### Open Questions
- Should IO acknowledge receipt of the attachment before processing?
- How to handle large files that exceed model context limits?

---

### How attachments reach the model
- **Images**: Pass as base64-encoded image content blocks in the messages API (Claude vision support). Format: `{ type: "image", source: { type: "base64", media_type, data } }`
- **Text files**: Inline the content in the user message (with filename header)
- **Binary/large files**: Store locally, pass a file path reference that tools can access

### Storage
- Temporary storage in `~/.io/attachments/` (or a configurable path)
- TTL-based cleanup (e.g., delete after 24h or after the conversation ends)
- Unique filename (UUID or hash) to avoid collisions

### Open Questions
- Max image resolution before downscaling? (Model limits apply)
- Multiple images in one message — pass all as separate content blocks?
- Should attachments be persisted in the chat history (SQLite) for replay, or ephemeral?

---

### How attachments flow to agents
When Michael says "fix this bug" with an attached screenshot, and IO delegates to a squad:
- The attachment must be passed as part of the task context to the squad agent
- The Copilot SDK's session message format needs to support image/file content blocks

### Open Questions
- Does `@github/copilot-sdk` support multi-modal content (image blocks) in messages? If not, what's the workaround?
  - Option A: Base64 in the message (if SDK supports content blocks)
  - Option B: Save to disk, pass file path as tool context, agent reads it
  - Option C: Upload to a temporary URL the agent can fetch
- Should squad agents be able to *produce* attachments back (e.g., generated diagrams)?
- How does this interact with squad instances (worktrees)? Are attachments scoped to the instance?

---

## Proposed Approach (for discussion)

1. **Attachment model**: `{ id, filename, mime_type, size, path, created_at }` stored in SQLite + file on disk
2. **TUI**: `/attach <path>` command adds to a pending attachments list, sent with next message
3. **Telegram**: Auto-detect photos/documents, download via Bot API, create attachment record
4. **Orchestrator**: Convert attachments to appropriate content blocks (image → base64 vision block, text → inline, other → path reference)
5. **Delegation**: Include attachment content/references in the task description passed to `delegateToAgent()`

---

## Acceptance Criteria (proposed)

- [ ] Users can attach images via TUI and have IO "see" them (vision)
- [ ] Telegram photos/documents are automatically processed as attachments
- [ ] Attachments are passed to Claude as appropriate content blocks
- [ ] When delegating to squads, attachments are included in task context
- [ ] Temporary storage with automatic cleanup
- [ ] Graceful handling of unsupported file types and size limits

michaeljolley:
Closes #268. Adds photo and document attachment support in Telegram bot with base64 blob pass-through to the Copilot SDK session.

### DIFF
--- a/src/copilot/orchestrator.ts
+++ b/src/copilot/orchestrator.ts
@@ -67,10 +67,18 @@ export type MessageSource =
 
 export type MessageCallback = (text: string, done: boolean) => void;
 
+export interface Attachment {
+  type: "blob";
+  data: string; // base64
+  mimeType: string;
+  displayName?: string;
+}
+
 interface QueuedMessage {
   prompt: string;
   source: MessageSource;
   callback: MessageCallback;
+  attachments?: Attachment[];
   resolve: () => void;
   reject: (err: Error) => void;
 }
@@ -438,6 +446,7 @@ function invalidateSession(): void {
 async function executeOnSession(
   prompt: string,
   callback: MessageCallback,
+  attachments?: Attachment[],
 ): Promise<string> {
   const session = await ensureOrchestratorSession();
 
@@ -449,7 +458,9 @@ async function executeOnSession(
   });
 
   try {
-    const result = await session.sendAndWait({ prompt }, SEND_TIMEOUT_MS);
+    const sendPayload: { prompt: string; attachments?: Attachment[] } = { prompt };
+    if (attachments && attachments.length > 0) sendPayload.attachments = attachments;
+    const result = await session.sendAndWait(sendPayload, SEND_TIMEOUT_MS);
     unsubDelta();
 
     const finalText = result?.data.content ?? accumulated;
@@ -513,7 +524,7 @@ async function processQueue(): Promise<void> {
 
       for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
         try {
-          const response = await executeOnSession(taggedPrompt, msg.callback);
+          const response = await executeOnSession(taggedPrompt, msg.callback, msg.attachments);
           logConversation("assistant", response, sourceLabel(msg.source));
           msg.resolve();
           lastError = undefined;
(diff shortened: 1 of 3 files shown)


michaeljolley (commented):
QA approved by WilyKat — size guards correct (document: pre+post, photo: post-only — acceptable since photo objects don't reliably expose file_size), error handling solid, SDK attachment format correctly wired through QueuedMessage→executeOnSession→sendAndWait.

Token security: URL is never logged, config-sourced not hardcoded, GitGuardian clean.

One non-blocking note: downloadTelegramFile doesn't check response.ok before calling response.arrayBuffer() — a failed download would encode an HTTP error body as base64 and pass it as attachment data. The outer try/catch prevents a crash, but a response.ok check would give a cleaner error. Follow-up fix, not blocking.

Both handlers correctly: largest photo selected (photos[photos.length-1]), caption as text, proper debounced-edit pattern, ack message deleted before streaming. CI green.

PR #284, opened 2026-05-25
fix: stream session events for delegate_to_teammate tasks

michaeljolley:
Fixes a bug where clicking Preview on non-lead squad agents showed 'Live' status but no activity.

**Root cause:** The delegate_to_teammate handler creates a task and runs it on the agent session, but never subscribes to session events via session.on(). Only the lead agent path (delegateToAgent) had this subscription.

**Fix:** Add recordTaskEvent subscription + cleanup in the delegate_to_teammate handler, mirroring the pattern in delegateToAgent.

### DIFF
--- a/src/copilot/agents.ts
+++ b/src/copilot/agents.ts
@@ -902,6 +902,21 @@ function buildAgentTools(squadSlug: string, isLead = false) {
               teammateAgent,
               task,
             );
+
+            recordTaskEvent(childTaskId, {
+              ts: Date.now(),
+              type: "task.start",
+              data: { taskId: childTaskId, agentKey: childAgentKey, description: task },
+            });
+            const unsubChild = session.on((event: SessionEvent) => {
+              if (!STREAM_EVENT_TYPES.has(event.type)) return;
+              recordTaskEvent(childTaskId, {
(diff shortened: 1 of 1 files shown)


michaeljolley (commented):
Changes requested by WilyKat: the catch exit path is missing unsubChild().

The delegateToAgent reference implementation uses a **finally block** to guarantee unsubscribe on all paths:
```ts
} finally {
  try { unsubscribe(); } catch { /* ignore */ }
}
```

The new delegateToTeammate code calls unsubChild() on success (line 943) and on timeout (line 939), but the inner catch block (line 946) has no unsubChild() call. Any exception thrown during sendWithIdleTimeout or result processing leaks the subscription.

Fix: wrap the unsubChild() calls in a finally block instead of duplicating them on each exit path, matching the delegateToAgent pattern exactly.

michaeljolley (commented):
QA approved by WilyKat — finally block ensures no subscription leak on any path. unsubChild?.() with optional chaining also correctly handles the edge case where session.on() throws before unsubChild is assigned. Pattern now matches delegateToAgent exactly. CI green.

Issue #315, opened 2026-05-26
Orchestrator chat fails


michaeljolley:
Occassionally when chatting with orchestrator, the response from it comes through as HTML

<img width="1359" height="665" alt="Image" src="https://github.com/user-attachments/assets/bd3d1819-abe0-4516-a64d-78c39923f0ab" />

michaeljolley:
Fixed in PR #316 (orchestrator HTML encoding normalization). Deployed in v0.41.0.

PR #316, opened 2026-05-26
fix: orchestrator chat HTML encoding issue (#315)

michaeljolley:
Fixes a bug where chat responses containing HTML-like content would sometimes render as literal escaped entities.

## Changes
- Add normalizeHtmlEncoding() helper to web/src/lib/markdown.ts
- Detects pre-escaped HTML entities and unescapes them once
- Integrated into renderMarkdown() at the start of processing
- Ensures proper rendering regardless of upstream encoding

## Testing
- All 284 backend tests pass
- All 49 web tests pass
- No new test failures

## Context
Issue #315 reported that orchestrator chat responses occasionally render as raw HTML instead of being properly parsed. Root cause: SDK or upstream components may pre-escape HTML entities, causing renderMarkdown to treat them as literal text instead of HTML tags. This fix normalizes encoding before markdown processing.

### DIFF
--- a/web/src/lib/markdown.ts
+++ b/web/src/lib/markdown.ts
@@ -40,6 +40,30 @@ hljs.registerLanguage('yml', yaml)
 hljs.registerLanguage('markdown', markdown)
 hljs.registerLanguage('md', markdown)
 
+/**
+ * Detects if a string appears to be HTML entities and unescapes them if needed.
+ * This handles cases where content may have been HTML-encoded multiple times.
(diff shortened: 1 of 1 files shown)


michaeljolley (commented):
Approved by QA — HTML encoding normalization logic is correct. Detection regex, single-unescape, and HTML-tag validation prevent over-unescaping legitimate content. Build clean, 49/49 web tests pass.

Issue #417, opened 2026-05-29
Bug: duplicate ellipsis shown while orchestrator is typing (static + animated)


michaeljolley:
## Description

While the orchestrator is processing and no response content has arrived yet, two ellipsis indicators are shown simultaneously — a static text ellipsis and an animated ellipsis. This is redundant.

## Fix

Remove the static ellipsis box. Keep only the animated ellipsis.

michaeljolley:
Reopening this issue because it isn't complete. It appears that we removed the ellipsis and left the animated ellipsis, but the bubble that held the removed ellipsis still exists. See this screenshot.

<img width="384" height="145" alt="Image" src="https://github.com/user-attachments/assets/b2908319-6581-4533-803d-8d400ee7fb25" />

michaeljolley:
Needs more scope before work starts: please confirm which component renders the static and animated ellipses so the redundant UI can be removed without affecting other typing states.

PR #432, opened 2026-05-29
fix: remove duplicate streaming ellipsis in chat

michaeljolley:
Removes the static fallback ellipsis while assistant messages are streaming so only the animated indicator remains.\n\nCloses #417


michaeljolley (commented):
LGTM — approved by Paul. Removed the static ellipsis fallback during streaming so only the animated indicator remains in both chat views.

Issue #420, opened 2026-05-29
Open all agent-generated links in a new window/tab


michaeljolley:
## Description

Any links generated by agents (in chat messages, feed posts, etc.) should open in a new window/tab (`target="_blank"`) rather than navigating away from the IO app.

PR #433, opened 2026-05-29
feat: open markdown links in new tabs

michaeljolley:
Uses the shared markdown renderer to add  and  to generated links so agent-produced content stays in the app.\n\nCloses #420


michaeljolley (commented):
LGTM — approved by Paul. Shared markdown rendering now forces external links to open in a new tab across chat, feed, and previews.

PR #476, opened 2026-06-18
chore(orchestrator): enforce instance isolation in system prompt

michaeljolley:
Adds an Instance Isolation section to the orchestrator system prompt enforcing instance creation and cleanup for code task delegations. Rules added: must call squad_instance_create before squad_delegate for code changes; exceptions for read-only tasks; limit of 3 instances per squad and cleanup policy; destroy instances after completion; branch naming using squad workflow. Aligns with IO squad workflow and operations SOP to prevent cross-task contamination and enforce safe parallelism.

### DIFF
--- a/src/copilot/system-message.ts
+++ b/src/copilot/system-message.ts
@@ -49,11 +49,11 @@ Violation of this rule is a HARD FAILURE — no exceptions, no workarounds, no "
 
 ## Instance Isolation
 Instance Isolation Rules:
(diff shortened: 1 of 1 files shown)


michaeljolley (commented):
LGTM — QA review by Murdock

michaeljolley (commented):
LGTM — approved by Hannibal

